PR #24, opened 2017-08-08
fix(FEC-6852): remove none functional components from ui

dvh91:


dan-ziv:
@dvh91 can u add description for the pr?
I didn't really understand why u comment those components?
If those components removed by default why not to delete them?

10x

dvh91:
@dan-ziv These components should be removed by default but still will be in use in phase 2 so I'm keeping the code but don't use those components.

PR #42, opened 2017-09-06
fix(FEC-7079): android fullscreen with custom ui

dvh91:


src/components/fullscreen/fullscreen.js
```diff
@@ -98,7 +98,10 @@ class FullscreenControl extends BaseComponent {
    * @memberof FullscreenControl
    */
   enterFullscreen(): void {
-    if (this.props.isMobile) {
+    var ua = navigator.userAgent.toLowerCase();
+    var isIOS = /ipad|iphone|ipod/.test(ua) && !window.MSStream;
```
dan-ziv: use the util of the player:
this.player.env

I think u can do:
`this.player.env.os.name === 'iOS'`

read more here:
https://github.com/faisalman/ua-parser-js

dan-ziv: use the util of the player:
this.player.env

I think u can do:
`this.player.env.os.name === 'iOS'`

read more here:
https://github.com/faisalman/ua-parser-js

PR #44, opened 2017-09-06
fix(FEC-6914): open menu while other menu open

dvh91:


OrenMe:
@dvh91 no other way beside setting explicit checks on play-overlay or dropdown-menu?

PR #48, opened 2017-09-12
fix(FEC-7074): removal of skip button and learn more button

dvh91:


src/ui-presets/ads.js
```diff
@@ -19,32 +16,10 @@ import BottomBar from '../components/bottom-bar';
  * @returns {HTMLElement} player ui tree
  */
 export default function adsUI(props: any): React$Element<any> {
-  var useStyledLinearAds = false;
-
-  try {
-    useStyledLinearAds = props.player.config.plugins.ima.adsRenderingSettings.useStyledLinearAds;
```
dan-ziv: `useStyledLinearAds = props.player.config.plugins.ima.adsRenderingSettings.useStyledLinearAds && !props.player.env.device.type`

dan-ziv: //TODO: When we will implement our ads manager this will come relevant
useCustomSkipButton = false;

PR #55, opened 2017-09-17
patch: attaching video element to DOM and refactoring around this change

dvh91:


src/components/keyboard/keyboard.js
```diff
@@ -16,7 +16,7 @@ class KeyboardControl extends BaseComponent {
   constructor(obj: Object) {
     super({name: 'Keyboard', player: obj.player});
 
-    this.player.getView().parentElement.onkeydown = (e) => {
+    this.player.getView().onkeydown = (e) => {
```
OrenMe: why not do it on the main UI container?

PR #60, opened 2017-09-27
feat: use off caption from playkit

OrenMe:

PlayKit creates a pseudo off text track when there are text tracks available.
This is for connivance only and it internally just tells PlayKit to call hideTextTrack, but it supports the model of emitting the text track change with the selected track where in case of off option it will properly add it to the event payload instead of just throwing null.
Depends on https://github.com/kaltura/playkit-js/pull/118

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/language/language.js
+++ b/src/components/language/language.js
@@ -118,20 +118,14 @@ class LanguageControl extends BaseComponent {
   }
 
   /**
-   * check if option is 'off', if it does- hideTextTrack called.
-   * otherwise, selecting the given text track
+   * Select the given text track
    *
-   * @param {(Object | string)} textTrack - text track or 'off' string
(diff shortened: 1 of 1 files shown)


src/components/language/language.js
```diff
@@ -118,20 +118,14 @@ class LanguageControl extends BaseComponent {
...
-   * check if option is 'off', if it does- hideTextTrack called.
-   * otherwise, selecting the given text track
+   * Select the given text track
    *
-   * @param {(Object | string)} textTrack - text track or 'off' string
+   * @param {(Object | string)} textTrack - text track
    * @returns {void}
    * @memberof LanguageControl
    */
   onCaptionsChange(textTrack: Object | string): void {
```
dan-ziv: Now the param type is only Object

PR #65, opened 2017-10-02
fix(FEC-7164): mobile seekbar on tap event handler

dvh91:


dan-ziv:
@dvh91 Can u resolve conflicts please?

PR #69, opened 2017-10-03
feat: remove call to external google font for better load time

OrenMe:

Remove the usage of external google font in order to get better load time.
Currently the request and processing may add up to 0.5s

### CheckLists

- [ ] Docs have been updated


dvh91:
@OrenMe - What's the replacement? This font is used for the default skin.

OrenMe:
@dvh91 no replacement ATM. Do we have another way to bundle font locally with UI?

dvh91:
@OrenMe We can use webfont loader for async load of the font so it won't block rendering.
https://github.com/typekit/webfontloader
Otherwise, I would advice Hillel

OrenMe:
But this will create a possible flash - let's consult with UX/UI and see if we can use default fonts instead.
Addning async font loading library will also defer the purpose of keeping us small as possible

PR #77, opened 2017-10-13
feat: Implementing css modules

dvh91:

Implementing css modules

### CheckLists

- [ ] Docs have been updated


OrenMe:
LGTM
@yairans, @dan-ziv, @odedhutzler - CSS modules enable encapsulation of CSS.
On our case we currently use it mainly to avoid leaking CSS selectors outside of player UI and this is to ensure that we don't have to think about it while implementing as the build step with webpack adds a prefix to any class we add.
Just be sure to import and use the scss in any component you create and don't inline classes or any CSS. 
@itaykinnrot - FYI

PR #116, opened 2017-11-01
feat(FEC-7140): seekbar thumbs sprite from config

dvh91:

Seekbar thumbs should be agnostic and get stripe image url, slices count and slice width from config.
config sample of kaltura-player-js:
```
let config = {
 ...
 ui: {
      SeekBarPlaybackContainer: {
        thumbsSprite: 'http://cdntesting.qa.mkaltura.com/p/1091/sp/109100/thumbnail/entry_id/0_wifqaipd/version/100042/height/90/width/160/vid_slices/200',
        thumbsWidth: 160,
        thumbsSlices: 200
      }
    }
}
```

### CheckLists

- [ ] Docs have been updated


src/components/base.js
```diff
@@ -30,6 +30,20 @@ class BaseComponent extends Component {
...
   }
+
+  /**
+   * Gets config param value
+   * @param {string} property property name
+   * @returns {string} config value
+   */
+  getConfig(property: string): ?string {
+    try {
+      return this.config.components[this.name][property];
```
OrenMe: isn't it better to solve the config object on init then to try-catch on each access?
IMO - better init the config in the constructor once with validation that it exist and then have the getConfig just return the property
```javascript
   constructor(obj?: Object = { config: {} }) {
     ...
     this.config = (obj.config.components && obj.config.components[this.name]) || {};
      ...
    }
```
```javascript
getConfig(property: string): ?string {
   return this.config[property];
}
```

this way a component will also carry only its config and not all components config

OrenMe:
@dvh91 were you able to look at my previous comment? how can we enable config changes on the fly?

dvh91:
@OrenMe Yes. I finalise the solution now. basically the solution is to expose new uimanager 'updateConfig' method save config obj in the store. then consume this config when needed in components. once the config will be changed in store, the component that use the relevant config will be rerendered.

dvh91:
`kaltura-player.js` required changes:
```
  constructor(player: Player, targetId: string, config: Object, uiConfig: ?Object) {
    ...
    return {
      ...
      setConfig: this.setConfig.bind(this)
    }
  }

  setConfig(config: Object, componentAlias?: string) {
    this._uiManager.setConfig(config, componentAlias);
  }
```

src/components/seekbar/seekbar.js
```diff
@@ -275,9 +277,12 @@ class SeekBarControl extends Component {
   getTimeBubbleOffset(): number {
     if (this._timeBubbleElement) {
       let leftOffset = (this.state.virtualTime / this.props.duration * this._seekBarElement.clientWidth) - (this._timeBubbleElement.clientWidth / 2);
-      if (leftOffset < 0) return 0;
-      else if (leftOffset > this._seekBarElement.clientWidth - this._timeBubbleElement.clientWidth) return (this._seekBarElement.clientWidth - this._timeBubbleElement.clientWidth);
-      else return leftOffset;
+      if (leftOffset < 0)
```
OrenMe: add  `{}` in all conditions

PR #117, opened 2017-11-01
feat: unmute indication for autoplay with volume failed

dvh91:

Adding unmute indication in case player tried to autoplay with volume and failed.

### CheckLists

- [ ] Docs have been updated


OrenMe:
With ads the mute indication doesn't appear - need to handle and also make sure the new indication doesn't interfere with the top bar.
And also, seem like there's a regression as an exception is thrown when trying to use volume bar.

OrenMe:
@dan-ziv @dvh91 please sync as this is causing the following behviour - when starting muted autoplay on Safari and clicking on the icon the ad pauses and the click handler goes crazy sometimes.

PR #132, opened 2017-11-11
feat: expose logger API

OrenMe:

set to ERROR level by default
expose API to set and get log level and get log level enums
expose convenience method to set logger level via global window param window.PLAYKIT_LOG_LEVEL
update to JS-Logger V1.4

### CheckLists

- [ ] Docs have been updated


src/ui-manager.js
```diff
@@ -48,6 +50,9 @@ class UIManager {
   constructor(player: Player, config: Object) {
     this.player = player;
     this.config = config;
+    if (config.logLevel && this.LogLevel[config.logLevel]) {
+      setLogLevel(this.LogLevel[config.logLevel]);
```
zoharbabin: @OrenMe @dan-ziv - this line seems to be faulty. It is not possible to set config level now, as this line always evaluates to undefined (LogLevel is an ENUM that contains objects of {value: 1, name: "DEBUG"}, etc.. It does not contain values).

zoharbabin: debug-issue

PR #133, opened 2017-11-15
fix(FEC-7368): when fallback happen - no loading spinner between thumbnail and ad playback

odedhutzler:

Added a "afterFirstPlay" state to the loading component.

### CheckLists

- [ ] Docs have been updated


src/components/loading/loading.js
```diff
@@ -112,6 +112,15 @@ class Loading extends BaseComponent {
     this.player.addEventListener(this.player.Event.ALL_ADS_COMPLETED, () => {
       this.props.updateLoadingSpinnerState(false);
     });
+
+    this.player.addEventListener(this.player.Event.FIRST_PLAY, () => {
+      this.setState({afterFirstPlay: true});
+      this.props.updateLoadingSpinnerState(false);
```
dan-ziv: Don't hide the spinner here since we didn't start toplay yet,
the PLAYER_STATE_CHANGED event handler will hide it on "playing"

src/components/loading/loading.js
```diff
@@ -89,10 +89,10 @@ class Loading extends BaseComponent {
     }
 
     this.player.addEventListener(this.player.Event.PLAYER_STATE_CHANGED, e => {
-      if (e.payload.newState.type === 'idle' || e.payload.newState.type === 'playing' || e.payload.newState.type === 'paused') {
+      if ((e.payload.newState.type === 'idle' || e.payload.newState.type === 'playing' || e.payload.newState.type === 'paused') && this.state.afterFirstPlay) {
         this.props.updateLoadingSpinnerState(false);
       }
-      else {
+      else if (this.state.afterFirstPlay) {
```
dan-ziv: Why we need this `if` statement?

odedhutzler: Because if we won't have it - we will have a spinner after first play
I want to ignore all player_state_changed events before first play

src/components/loading/loading.js
odedhutzler: fixed

src/components/loading/loading.js
dan-ziv: so first check this state, don't use it twice
`this.player.addEventListener(this.player.Event.PLAYER_STATE_CHANGED, e => {`
 `if (!this.state.afterFirstPlay) return;`

and now u can give up both

PR #144, opened 2017-11-27
feat(FEC-7472): error handling ui

odedhutzler:

Created a preset for the error UI.
- Retry Button.
- Copy to clipboard the error session id

### CheckLists

- [ ] Docs have been updated


src/components/engine-connector/engine-connector.js
```diff
@@ -157,6 +157,11 @@ class EngineConnector extends BaseComponent {
       this.props.updateAdSkippableState(e.payload.ad.getAdSkippableState());
     });
 
+    this.player.addEventListener(this.player.Event.ERROR, e => {
+      if (e.payload){
+        this.props.updateError(e.payload);
```
dan-ziv: I think better name is `updateHasError`

src/reducers/engine.js
```diff
@@ -49,11 +50,18 @@ export const initialState = {
     currentTime: 0,
     duration: 0
   },
-  adUrl: ''
+  adUrl: '',
+  error: ''
```
dan-ziv: change to `hasError` since it's more meaningful

src/components/error-overlay/error-overlay.js
```diff
@@ -0,0 +1,155 @@
...
+ * @returns {Object} - mapped state to this component
+ */
+const mapStateToProps = state => ({
+  error: state.engine.error
+});
+
+const defaultError = "Something went wrong";
+const defaultSessionText = "Session ID ";
+const buttonText = "Retry";
+
```
dan-ziv: Use capital letters only for const vars

dan-ziv: Move strings to i18n

src/components/error-overlay/error-overlay.js
```diff
@@ -0,0 +1,155 @@
...
+  /**
+   * copy input text based on input element.
+   * on success, set success internal component state for 2 seconds
+   * @returns {void}
+   * @memberof ShareOverlay
+   */
+  copyError(): void {
+    let selection = window.getSelection();
+    let dataEl = document.querySelector("." + style.errorSession);
```
dan-ziv: don't refer a dom element like this when u can use preact for this.
see ref attribute [https://reactjs.org/docs/refs-and-the-dom.html](url)

src/components/error-overlay/error-overlay.js
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  /**
+   * render main state
+   *
+   * @returns {React$Element} - main state element
+   * @memberof ErrorOverlay
+   */
+  render(): React$Element<any> | void {
+    let copyUrlClasses = [style.btnCopyUrl].join(' ');
```
dan-ziv: In general think on split things down here to ui components for reuse and readability.
In addition, I think that too much definitions are inline here - can we export it to css files?
What about ids to dom elements? do we really need it?

src/components/share-overlay/_share-overlay.scss
```diff
@@ -58,7 +58,7 @@
         position: absolute;
         width: 32px;
       }
-      .check-icon {
+      .icon-check {
```
dan-ziv: Why those changes?

src/components/overlay/overlay.js
```diff
@@ -42,6 +42,24 @@ class Overlay extends Component {
...
   }
 
+  /**
+   * closeButton
+   * @param {any} props - props
+   * @returns {React$Element | void} close button element
+   * @memberof Overlay
+   */
+  renderCloseButton(props: any): React$Element<any> | void {
+    if (!props.permanent) {
```
dan-ziv: What is `props.permanent`?

src/components/overlay/overlay.js
```diff
@@ -42,6 +42,24 @@ class Overlay extends Component {
...
+      return (
+        <Localizer>
+          <a onClick={() => props.onClose()} aria-label={<Text id='core.close'/>} className={style.closeOverlay}><Icon
+            type={IconType.Close}/></a>
+        </Localizer>)
+    } else {
```
dan-ziv: no need the `else`
just `return undefiend;`

src/player-gui.js
```diff
@@ -12,7 +12,8 @@ const mapStateToProps = state => ({
     shell: state.shell,
     engine: {
       adBreak: state.engine.adBreak,
-      isLive: state.engine.isLive
+      isLive: state.engine.isLive,
+      error: state.engine.error
```
dan-ziv: change name to `hasError`

src/reducers/engine.js
```diff
@@ -20,7 +20,8 @@ export const types = {
   UPDATE_AD_URL: 'engine/UPDATE_AD_URL',
   UPDATE_PLAYER_POSTER: 'engine/UPDATE_PLATER_POSTER',
   UPDATE_IS_LIVE: 'engine/UPDATE_IS_LIVE',
-  UPDATE_IS_DVR: 'engine/UPDATE_IS_DVR'
+  UPDATE_IS_DVR: 'engine/UPDATE_IS_DVR',
+  UPDATE_ERROR: 'engine/ERROR'
```
dan-ziv: `UPDATE_HAS_ERROR: 'engine/HAS_ERROR'`

src/ui-presets/error.js
```diff
@@ -0,0 +1,19 @@
...
+ * Error ui interface
+ *
+ * @export
+ * @param {*} props component props
+ * @returns {React$Element} player ui tree
+ */
+export default function errorUI(props: any): React$Element<any> {
+  return (
+    <div className={style.playbackGuiWWrapper}>
+      <ErrorOverlay player={props.player}/>
```
dan-ziv: Why all the elements are in the OverlayError components? This is very hard to reuse like this.
As I see it the overlay should only define the overlay itself (look on the overlay-play for reference) and the other elements should be map to react components underneath

src/components/error-overlay/_error-overlay.scss
```diff
@@ -0,0 +1,79 @@
+@import '~styles/variables';
```
dan-ziv: beautify this file

src/components/share-overlay/_share-overlay.scss
odedhutzler: because it either changing it here, or changing it in the share component

dan-ziv: Why to change it at all?

odedhutzler: that's the structure of all the icons in the icon component. 
<i className={[style.icon, style.iconCheck].join(' ')}/>);

dan-ziv:
@odedhutzler 
SVGs should be defined in `_icons.scss` and` icon.js` and not inline

src/components/overlay/overlay.js
odedhutzler: it is a new prop i am passing, to determine if the window is permanent (meaning it can't be closed) - so no close button is needed

src/player-gui.js
odedhutzler: I think a state named 'error' is pretty clear.

AnnaAleksandrowicz:
My two cents: I would go with a state that is more precise than "error" or "haserror" - if at all possible- to make it easier to the developer to know what is the problem.

odedhutzler:
@AnnaAleksandrowicz - thanks! The purpose of this state member is to indicate that there is an error. Currently there are 9 categories of errors which are all mapped into the same error page. I guess that if we decide to split them in the future, we can use different states for different errors.

AnnaAleksandrowicz:
Got it :)

dan-ziv:
@OrenMe @odedhutzler 
Few things:
1. Copy to clipboard should be a separate component with a very specific logic and not part of the error overlay.
We will need to use this component in some other places in the future (see Hillel's mockups).
2. entry ID is something very Kaltura Player oriented and we discuss that the UI is not bind to any player. Maybe pass an handler is better than an entry ID. 
3. There's no such thing errorOverlay - there's an error ui which contains the overlay component. So if u pass any config to it, the config key should called errorUI.

PR #152, opened 2017-11-29
fix: focus player shell on on mouse hover

dan-ziv:

Focus the player on mouse hover to allow control the player with keyboard/tabbing without the need of clicking it.

### CheckLists

- [ ] Docs have been updated


OrenMe:
@dan-ziv why do we need this? the site should take care of tab integration order, and there would be no need to handle this via mouse.
Also, there's no sense to add mouse to accessibility :-)

PR #156, opened 2017-12-10
fix(FEC-7368): no spinner between ads and spinner doesn't stop on autoplay on native adapter

odedhutzler:
Added a state (afterPlayingEvent) the checks if the entry started playing, it listens to the PLAYING event.
PLAYER_STATE_CHANGED events are ignored until 'afterPlayingEvent' state is set to true.

@dan-ziv, i checked with/without autoplay and with/without ads. It seems to work, but I remember you have some notes about this fix.

### CheckLists

- [ ] Docs have been updated


src/components/loading/loading.js
```diff
@@ -89,6 +90,9 @@ class Loading extends BaseComponent {
     }
 
     this.player.addEventListener(this.player.Event.PLAYER_STATE_CHANGED, e => {
+      if (!this.state.afterPlayingEvent) {
```
OrenMe: I think we are starting to have duplicate logic here, and maybe it's better to consider moving to playback events based state only and not state change event.
You already see that we both register to playing event a few line below and do the same action there and also in here in state change to playing.

PR #158, opened 2017-12-13
fix(FEC-7451): video does not play after pressing play

odedhutzler:
Removed duplicate call to the ready promise.
It seems calling play would be enough in case of preload: 'auto'. if the video is not loaded yet - then it will be handled in the player.js

### CheckLists

- [ ] Docs have been updated


odedhutzler:
@OrenMe @dan-ziv  besides the 'quick click refresh' issue, it seems to be ok. (tested on mobile & mac with autoplay true/false & preload:none/auto).

src/components/pre-playback-play-overlay/pre-playback-play-overlay.js
```diff
@@ -94,25 +94,14 @@ class PrePlaybackPlayOverlay extends BaseComponent {
...
-        if (this.player.config.playback.preload === "auto" && !this.player.config.plugins.ima) {
-          this.player.ready().then(resolve);
-        } else {
-          resolve();
-        }
-      } catch (e) {
-        reject(e);
-      }
-    }).then(() => {
+    try {
```
OrenMe: @odedhutzler The try catch was there to protect against the nested player accessor I think, please check if still required.
Otherwise, LGTM.

src/components/pre-playback-play-overlay/pre-playback-play-overlay.js
```diff
@@ -126,6 +114,7 @@ class PrePlaybackPlayOverlay extends BaseComponent {
     if ((!props.isEnded && !props.prePlayback) || (!props.isEnded && this.autoplay)) {
       return undefined;
     }
+    let playerIsReady = this.player.config.playback.preload === "auto" ? this.state.isPlayerReady : true;
```
dan-ziv: use const
why state name is isPlayerReady and var name is  playerIsReady?

odedhutzler: changed to const + naming, thanks

PR #207, opened 2018-03-26
fix(FEC-8079): hide UI on player RESET event

dan-ziv:

Add isStopped state in the engine reducer and do not render the PlayerGUI component when player is stopped! (we still want to render the other components, like VideoPlayer for example, since the video element placed there)

### CheckLists

- [ ] Docs have been updated


odedhutzler:
so from what i understand here, we will stay with the old ui (and spinner) until the stop flag is turned false (meaning the player reseted?)

PR #210, opened 2018-04-08
fix(FEC-8099): remove the seek bar while ad

yairans:

the seek bar is unnecessary while ad, also it takes needed space - removed while ad

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/ui-presets/ads.js
+++ b/src/ui-presets/ads.js
@@ -51,12 +51,6 @@ export default function adsUI(props: any): ?React$Element<any> {
           {adsUiCustomization.skipButton ? <AdSkip player={props.player}/> : undefined}
         </div>
         <BottomBar>
-          <SeekBarAdsContainer
-            adBreak
-            showFramePreview
-            showTimeBubble
-            player={props.player}
-            playerContainer={props.playerContainer}/>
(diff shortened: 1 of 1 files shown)


src/ui-presets/ads.js
```diff
@@ -51,12 +51,6 @@ export default function adsUI(props: any): ?React$Element<any> {
           {adsUiCustomization.skipButton ? <AdSkip player={props.player}/> : undefined}
         </div>
         <BottomBar>
-          <SeekBarAdsContainer
```
dan-ziv: @OrenMe @yairans 
Why not to do it configurable?

dan-ziv: Or even better,
Why not to check in the seekbar component if the uiElements appear in IMA config and just in that case to remove the seekbar?

OrenMe: IMO you want the behaviour to be consistent and not dependent on other 3rd party factors.
Also, the seek bar and the time label act as the same here as both indicate time, and seek bar doesn't have it's usual seek behaviour, so in terms of functionality they do the same - and in our case the seekbar also takes needed space - so better to just remove it.

PR #211, opened 2018-04-08
feat(FEC-7330): enable setting custom track labels by app

odedhutzler:

added label support in ui, if there is label field in video track is empty - it will fallback to height+'p'

### CheckLists

- [ ] Docs have been updated


src/components/settings/settings.js
```diff
@@ -134,7 +134,7 @@ class SettingsControl extends BaseComponent {
    * @memberof SettingsControl
    */
   getQualityOptionLabel(t: Object): string {
-    let resolution = t.height ? t.height + 'p' : undefined;
+    let resolution = t.label ? t.label : t.height ? t.height + 'p' : undefined;
```
yairans: * make it more readable
* change the `resolution` var to `label`

src/components/settings/settings.js
```diff
@@ -134,16 +134,16 @@ class SettingsControl extends BaseComponent {
...
     if (!this.props.qualityType) {
-      return resolution || mbs || 'N/A';
+      return qualityLabel || mbs || 'N/A';
     }
     else if (this.props.qualityType.toUpperCase() === 'MBS' && mbs) {
       return mbs;
     }
-    else if (this.props.qualityType.toUpperCase() === 'RESOLUTION' && resolution) {
+    else if (this.props.qualityType.toUpperCase() === 'RESOLUTION' && qualityLabel) {
       return t.height + 'p';
```
OrenMe: we already calculate the resolution label above, just like mbs, so need to return  `resolution`.

OrenMe: also, resolution was the proper name, as we calculate resolution or bitrate.

PR #212, opened 2018-04-08
fix(FEC-8107): fullscreen APIs on IOS devices

dan-ziv:

Fix full screen APIs logic on iOS devices.

### CheckLists

- [ ] Docs have been updated


src/components/fullscreen/fullscreen.js
```diff
@@ -167,7 +167,9 @@ class FullscreenControl extends BaseComponent {
    * @memberof FullscreenControl
    */
   exitFullscreen() {
-    if (typeof document.exitFullscreen === 'function') {
+    if (this.props.isMobile && this.player.env.os.name === 'iOS') {
```
OrenMe: don't we need this for tablet(iPad) as well? so in general iOS?

PR #218, opened 2018-04-11
fix(FEC-7836): seek doesn't work correctly on MAC safari on fullscreen

odedhutzler:

Safari fullscreen element has >0 offset. Chrome (for example) gives it offset 0.

I added isfullscreen() check and set the offset to 0.

tested on mac(chrome+safari), ipad, xiomi & samsung tab 3

### CheckLists

- [ ] Docs have been updated


src/components/seekbar/seekbar.js
```diff
@@ -270,6 +270,9 @@ class SeekBarControl extends Component {
       _y += element.offsetTop - element.scrollTop;
       element = element.offsetParent;
     }
+    if (this.props.player.isFullscreen()) {
+      _x = 0;
```
OrenMe: I think this is worth an inline comment, like 

> offset 0 is forced to handle browser compatibility issue

PR #223, opened 2018-04-18
fix(FEC-8032): when switching to iOS fullscreen in portrait captions displayed in the middle for few seconds

odedhutzler:
- when detecting iOS and full screen, removing the transition -60px because there is no bottom menu bar. 
- added environment classes to the player main class to indicate browser and device.
- added helper 'playkit' global var to the webpack build.

### CheckLists

- [ ] Docs have been updated


src/components/shell/shell.js
```diff
@@ -276,6 +277,7 @@ class Shell extends BaseComponent {
     if (this.props.metadataLoaded) playerClasses.push(style['state-' + this.props.currentState]);
     if (this.props.seekbarDraggingActive) playerClasses.push(style.hover);
     if (this.props.playerClientRect && this.props.playerClientRect.width <= 480) playerClasses.push(style.sizeSm);
+    if (this.props.fullscreen && this.player.env.os.name === 'iOS') playerClasses.push(style.iOSFullscreen);
```
OrenMe: Let's move this condition from between the size condition (row above and below).
Also, let's just add fullscreen state class, e.g. `if (this.props.fullscreen) playerClass.push(style.inFullscreen)` (or maybe just fullscreen)
and add user agent class, so:
```
let playerClasses = [style.player, style.skinDefault, this.player.env.os.name, this.player.env.browser.name];
playerClasses.push(props.playerClasses);
```

Probably need to add the `playkit` prefix to these new classes.

ref in V2: https://github.com/kaltura/mwEmbed/blob/ba045a91abdee7d0761ba543f5068b4335c04c9f/modules/EmbedPlayer/resources/mw.PlayerLayoutBuilder.js#L152-L177

src/components/shell/shell.js
```diff
@@ -265,7 +266,9 @@ class Shell extends BaseComponent {
    * @memberof Shell
    */
   render(props: any): React$Element<any> {
-    let playerClasses = [style.player, style.skinDefault];
+    let playerClasses = [style.player, style.skinDefault,
+      `playkit-${this.player.env.os.name.replace(/ /g, "_")}`,
```
OrenMe: We can add the prefix as global in webpack.config and add this as token here.
See here how we use it with _webpack.DefinePlugin_ 
We define it here: https://github.com/kaltura/playkit-js-ui/blob/afe561f9f7953e7f64208f3fc9cf2f0edbab3c80/webpack.config.js#L9-L13

and use it here: 
https://github.com/kaltura/playkit-js-ui/blob/afe561f9f7953e7f64208f3fc9cf2f0edbab3c80/src/index.js#L23-L24

so you can add another define in webpack.config ➡️ `const CSS_MODULE_PREFIX = 'playkit'` and use it in webpack config in the define plugin and also in the style loader config where we define the [localIdentName](https://github.com/kaltura/playkit-js-ui/blob/master/webpack.config.js#L69)

webpack.config.js
```diff
@@ -4,11 +4,13 @@ const webpack = require("webpack");
 const path = require("path");
 const PROD = (process.env.NODE_ENV === 'production');
 const packageData = require("./package.json");
+const CSS_MODULE_PREFIX = "playkit-";
```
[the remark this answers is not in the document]
odedhutzler: done + beautify

OrenMe:
@dan-ziv please re-review following fixes

src/components/shell/shell.js
```diff
@@ -7,6 +7,7 @@ import {bindActions} from '../../utils/bind-actions';
 import {actions} from '../../reducers/shell';
 import {KeyMap} from "../../utils/key-map";
 
+declare var __CSS_MODULE_PREFIX__: string;
```
dan-ziv: Move this declaration to the ui manager

src/components/shell/shell.js
```diff
@@ -265,7 +267,9 @@ class Shell extends BaseComponent {
    * @memberof Shell
    */
   render(props: any): React$Element<any> {
-    let playerClasses = [style.player, style.skinDefault];
+    let playerClasses = [style.player, style.skinDefault,
+      __CSS_MODULE_PREFIX__ + '-' + this.player.env.os.name.replace(/ /g, "_"),
+      __CSS_MODULE_PREFIX__ + '-' + this.player.env.browser.name.replace(/ /g, "_")];
```
dan-ziv: No need to re calculate it on every time render has been called. We can calculate it one time in the constructor and use it here.
```js
this._osClass = `${__CSS_MODULE_PREFIX__}-${this.player.env.os.name.replace(/ /g, '-')}`;
this._browserClass = `${__CSS_MODULE_PREFIX__}-${this.player.env.browser.name.replace(/ /g, '-')}`;

```
One more thing is to change the replacer string to '-' instead of '_' since this is our convention in css class names.

dan-ziv:
@odedhutzler update description

src/components/shell/shell.js
odedhutzler: done

PR #229, opened 2018-04-24
feat(FEC-8139): iOS in-browser fullscreen

dan-ziv:

Gives a user the ability to choose for his player an in-browser fullscreen experience on iOS devices which will replace the native fullscreen of the AV player.

user will need to configure it via the UI:

```js
ui: {
  components: {
    fullscreen: {
       inBrowserFullscreenForIOS: true
     }
  }
}
```

### CheckLists

- [ ] Docs have been updated


src/components/fullscreen/_fullscreen.scss
```diff
@@ -1,18 +1,26 @@
-.control-button-container.control-fullscreen {
-  .control-button {
-    transition: 100ms transform;
-    transform: scale(1);
+.in-browser-fullscreen-mode {
+  width: auto !important;
```
OrenMe: Need to also apply left/top 0 and z-index to max in case other elements have higher position

src/components/fullscreen/fullscreen.js
```diff
@@ -168,7 +200,11 @@ class FullscreenControl extends BaseComponent {
    */
   exitFullscreen() {
     if (this.player.env.os.name === 'iOS') {
-      this.player.getVideoElement().webkitExitFullScreen();
+      if (this.props.inBrowserFullscreenForIOSDevices) {
```
OrenMe: inBrowserFullscreenForIOS? devices is redundant

src/components/fullscreen/fullscreen.js
```diff
@@ -15,8 +15,10 @@ import {default as Icon, IconType} from '../icon';
  * @returns {Object} - mapped state to this component
  */
 const mapStateToProps = state => ({
+  containerId: state.config.containerId,
```
OrenMe: can be applied on the targetId as well, so this is not necessary

PR #230, opened 2018-04-24
fix(FEC-8146): loading spinner disappear too early when overlay ad

dan-ziv:

Save ad linear state and don't remove the loading spinner in case of a non linear ad.

### CheckLists

- [ ] Docs have been updated


src/components/loading/loading.js
```diff
@@ -14,7 +14,8 @@ import BaseComponent from '../base';
 const mapStateToProps = state => ({
   show: state.loading.show,
   isMobile: state.shell.isMobile,
-  adBreak: state.engine.adBreak
+  adBreak: state.engine.adBreak,
+  adIsLinear: state.engine.adIsLinear
```
yairans: I would rename to `isLinearAd`

PR #247, opened 2018-06-24
fix:add color white on the span itself

itaykinnrot:
add color to span


src/components/unmute-indication/_unmute-indication.scss
```diff
@@ -1,3 +1,5 @@
+$color: 'white';
```
dan-ziv: no need to define new scss var for white, this is already defined in the _variables.scss

src/components/unmute-indication/_unmute-indication.scss
```diff
@@ -6,7 +8,7 @@
   z-index: 15;
 
   a {
-    color: white;
+    color: $color;
```
dan-ziv: `color: $white;`

src/components/unmute-indication/_unmute-indication.scss
```diff
@@ -34,6 +36,7 @@
     opacity: 1;
     transition: transform 100ms, opacity 100ms;
     display: inline-block;
+    color: $color;
```
dan-ziv: `color: $white;`

src/components/unmute-indication/_unmute-indication.scss
itaykinnrot: its more for the color of the unmute button

dan-ziv: the color is white, so until this will be changed, no need for another definition IMO

itaykinnrot: @OrenMe ?

OrenMe: @dan-ziv as it is used in multiple places and we will need to support theming it's better to start putting such things as settings attribute - although it is not connected to anything ATM.
Also - setting a css param called white and then using it instead has no value over inlining it.

OrenMe: @dan-ziv if there's no other issue please remove the change request. thanks.

PR #253, opened 2018-07-11
feat(FEC-8395): Dont render zero or one option

itaykinnrot:

dont render options with one option (language and settings) + remove the isLive - playbackrate (move to core)
### CheckLists

- [X] Travis tests are passing (or test results are not worse than on master branch :))
- [ ] Docs have been updated


src/components/language/language.js
```diff
@@ -160,7 +160,8 @@ class LanguageControl extends BaseComponent {
    * @returns {React$Element} - component
    * @memberof LanguageControl
    */
-  renderAll(audioOptions: Array<Object>, textOptions: Array<Object>): React$Element<any> {
+  renderAll(audioOptions: Array<Object>, textOptions: Array<Object>): React$Element<any> | void {
+    if (audioOptions.length <=1 && textOptions.length <= 1) return undefined;
```
yairans: the `render` method already has condition of the length. edit it instead of adding a new condition.

yairans: what if audio has 1 the text has 2. do we want to render all?

yairans: beautify :)

src/components/settings/settings.js
```diff
@@ -205,7 +205,7 @@ class SettingsControl extends BaseComponent {
         });
     }
 
-    if (props.isLive && qualityOptions.length === 0) return undefined;
+    if (qualityOptions.length <= 1 && speedOptions.length <= 1) return undefined;
```
yairans: same Q as above. what is we have 1 quality and many sppeds

OrenMe: Yes, we need to do it per option menu - I think it's better to leave it here, per each option and remove the additional condition from the container and item components.

src/components/language/language.js
OrenMe: same as this: https://github.com/kaltura/playkit-js-ui/pull/253/files#r201715680

src/components/settings/settings.js
itaykinnrot: @yairans  if we have one quality and many speed - we want to render the button and the menu - the inner item will not be render since i added a check for it

PR #254, opened 2018-07-12
feat(FEC-8390, FEC-8246): support 608/708 captions

yairans:

* support text track of kind 'captions'.
* handle caption on `hover` state (need to support upper captions) with 2 ways: 
    * Native caption (safari or by configuration): by **resizing** the subtitle container dynamically  
    * Non-native captions(not-safari or  by configuration) : by **moving** the subtitle container dynamically (to avoid caption size change)

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/language/language.js
+++ b/src/components/language/language.js
@@ -176,7 +176,7 @@ class LanguageControl extends BaseComponent {
         </Localizer>
         {!this.state.smartContainerOpen || this.state.cvaaOverlay ? undefined :
           <SmartContainer title='Language' onClose={() => this.onControlButtonClick()}>
-            {audioOptions.length === 0 ? undefined :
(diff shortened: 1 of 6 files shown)


src/components/language/language.js
```diff
@@ -234,7 +234,7 @@ class LanguageControl extends BaseComponent {
     const audioOptions = props.audioTracks
       .filter(t => t.label || t.language)
       .map(t => ({label: t.label || t.language, active: t.active, value: t}));
-    const textOptions = props.textTracks.filter(t => t.kind === 'subtitles').map(t => ({
```
OrenMe: Why remove the filter? if anything then make it tighter - subtitles or captions.

yairans: we already filter the text tracks in the adapters.

PR #264, opened 2018-08-06
fix(FEC-8286): adding click indication to the state

odedhutzler:

when clicking on the unmute indication with the keyboard, it does not disappear.

when clicking with the mouse it does - that's because we have (mouse) click handlers in the UI that makes it happen. 

added a 'clicked' state that saves it, if it was clicked, and re-render accordingly.

### CheckLists

- [ ] Docs have been updated


odedhutzler:
@dan-ziv, removed the 'clicked' state hack and handled it like we handle a click in the shell component

PR #310, opened 2018-12-03
  fix(FEC-8738): mobile - when there are 2 players in a page, the settings from the 2nd are presented on the first 

odedhutzler:

changing portal element from id to class
changing the css accordingly

this is done so if there are 2 players, the portal will be created under the player specific player id.

### CheckLists

- [ ] Docs have been updated


src/styles/_shell.scss
```diff
@@ -42,14 +42,14 @@
       font-family: $font-family;
     }
   }
-  :global(#overlay-portal) {
+  :global(.overlay-portal) {
```
OrenMe: check if we can remove the global - as now we have the portal only under a player

src/components/smart-container/smart-container.js
```diff
@@ -36,7 +37,14 @@ const mapStateToProps = state => ({
  * </SmartContainer>
  * @extends {Component}
  */
-class SmartContainer extends Component {
+class SmartContainer extends BaseComponent {
+  /**
+   * @param {Object} obj an object
+   */
+  constructor(obj: Object) {
```
OrenMe: I think it is better to keep this a regular component and pass in the player id

src/styles/_shell.scss
odedhutzler: cannot remove it

src/components/smart-container/smart-container.js
odedhutzler: changed.

PR #322, opened 2018-12-24
fix(FEC-8784): can't set the volume to 1 or 0/mute via keyboard

shalommeoded:

Change the conditions of the key down\up to comply with use cases when the player.volume is not set lower than 0 and higher than 1.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/keyboard/keyboard.js
+++ b/src/components/keyboard/keyboard.js
@@ -37,12 +37,12 @@ export const KEYBOARD_DEFAULT_VOLUME_JUMP: number = 5;
   mapStateToProps,
   bindActions(Object.assign({}, shellActions, overlayIconActions))
 )
-/**
- * KeyboardControl component
- *
- * @class KeyboardControl
- * @extends {BaseComponent}
- */
+  /**
+   * KeyboardControl component
+   *
+   * @class KeyboardControl
+   * @extends {BaseComponent}
+   */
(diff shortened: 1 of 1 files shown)


src/components/keyboard/keyboard.js
```diff
@@ -111,18 +111,18 @@ class KeyboardControl extends BaseComponent {
       if (this.player.muted) {
         this.player.muted = false;
       }
-      if (newVolume <= 1) {
```
OrenMe: Just cap newVolume to 1 if it is bigger, no need to add this condition.
You can do it in line 109 where you calculate the newVolume.

src/components/keyboard/keyboard.js
```diff
@@ -111,18 +111,18 @@ class KeyboardControl extends BaseComponent {
...
+      if (newVolume <= 1 || newVolume > 1) {
         this.player.volume = newVolume;
         this.props.updateOverlayActionIcon([IconType.VolumeBase, IconType.VolumeWaves]);
         return {volume: this.player.volume};
       }
     },
     [KeyMap.DOWN]: () => {
       const newVolume = (Math.round(this.player.volume * 100) - KEYBOARD_DEFAULT_VOLUME_JUMP) / 100;
```
OrenMe: Same here, cap the volume to 0 and avoid the cumbersome conditions.

PR #336, opened 2019-01-22
fix(FEC-8852): closing pip window on Mac with Safari the video dose not return to the player

odedhutzler:

when closing the pip with the 'x' button on the window, safari fires a `PRESENTATION_MODE_CHANGED` event. 

listening to this event and checking if the player is in PIP mode or not.

### CheckLists

- [ ] Docs have been updated


odedhutzler:
@OrenMe 
because I don't manipulate safari events, I can listen to this event on playkit, do the same check there and fire leave / enter event.
(here it happens because we close the window from the PIP itself, not from the player)

PR #344, opened 2019-02-05
fix(FEC-8885): smart menu doesn't open if no label is supplied

odedhutzler:

in case there is not label supplied (e.g. field does not exist in the localization), the label will return as null. added protection.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/smart-container/smart-container-item.js
+++ b/src/components/smart-container/smart-container-item.js
@@ -21,7 +21,7 @@ class SmartContainerItem extends Component {
   render(props: any): React$Element<any> {
     return (
       <div className={[style.smartContainerItem, style.selectMenuItem].join(' ')}>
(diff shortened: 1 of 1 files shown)


src/components/smart-container/smart-container-item.js
```diff
@@ -21,7 +21,7 @@ class SmartContainerItem extends Component {
   render(props: any): React$Element<any> {
     return (
       <div className={[style.smartContainerItem, style.selectMenuItem].join(' ')}>
-        <label htmlFor={props.label.toLowerCase()}>
+        <label htmlFor={props.label && props.label.toLowerCase()}>
```
OrenMe: I suggest you do the check once before the return and then apply it in the jsx instead of repeating this twice

PR #350, opened 2019-02-20
fix(FEC-8903): when loading a player with display: none, the player size class is set to small

odedhutzler:

Besides checking for dimension on resize events, check for dimensions on `FIRST_PLAY` event.

this way, if the player is hidden before (with dispaly: none e.g) it will be rendered with it's real width and height when it is playing. 

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/shell/shell.js
+++ b/src/components/shell/shell.js
@@ -199,6 +199,7 @@ class Shell extends BaseComponent {
     this.props.updateIsMobile(!!this.player.env.device.type || this.props.forceTouchUI);
     this._onWindowResize();
     this.eventManager.listen(window, 'resize', () => this._onWindowResize());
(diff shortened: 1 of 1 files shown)


src/components/shell/shell.js
```diff
@@ -199,6 +199,7 @@ class Shell extends BaseComponent {
     this.props.updateIsMobile(!!this.player.env.device.type || this.props.forceTouchUI);
     this._onWindowResize();
     this.eventManager.listen(window, 'resize', () => this._onWindowResize());
+    this.eventManager.listen(this.player, this.player.Event.PLAYING, () => this._onWindowResize());
```
yairans: could be done in `FIRST_PLAYING` event? 
to avoid doing this each `PLAYING` event which triggered many times while playback.

odedhutzler: changed :)

PR #351, opened 2019-02-21
fix(FEC-8565): bottom bar is not clickable when the loading overlay is displayed

odedhutzler:

Reduced spinner overlay height, so the bottom bar will be clickable.

### CheckLists

- [ ] Docs have been updated


odedhutzler:
let remove the whole overlay and leave only the spinner

PR #395, opened 2019-07-08
Fec 8696 contrib presets

esakal:


flow-typed/types/ui-options.js
```diff
@@ -1,10 +1,12 @@
 // @flow
+
 declare type UIOptionsObject = {
   targetId: string,
   debugActions?: boolean,
   forceTouchUI?: boolean,
   log?: UILogConfigObject,
   components?: ComponentsConfig,
+  presetComponents?: PresetComponent[],
```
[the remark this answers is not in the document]
yairans: it already got default empty array from kaltura-player https://github.com/kaltura/kaltura-player-js/pull/253/files#diff-039b70137bb24fed2504a0093f23e583R13

src/ui-manager.js
```diff
@@ -169,13 +179,15 @@ class UIManager {
...
       const template = (
         <Provider store={this.store}>
-          <IntlProvider definition={this._translations[this._locale]}>
-            <Shell player={this.player}>
-              <EngineConnector player={this.player} />
-              <VideoPlayer player={this.player} />
-              <PlayerGUI uis={uis} player={this.player} playerContainer={this.container} />
-            </Shell>
-          </IntlProvider>
+          <ContainerProvider presetComponents={this._presetComponents}>
```
yairans:
```suggestion
          <ContainerProvider presetComponents={config.presetComponents}>
```
You can remove `this._presetComponents`, 
`config.presetComponents` typed as array as I mention above

src/components/container/container.js
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    const nextContainerComponents = {
+      appendedComponents: [],
+      positionedComponentMap: {}
+    };
+    const positionedComponentMap = nextContainerComponents.positionedComponentMap;
+    let hasPositionedComponents = false;
+
+    const presetComponents = [...presetsData.allPresets, ...(presetsData.specificPreset[targetPresetName] || [])];
```
yairans:
```suggestion
    const presetComponents = presetsData.allPresets.concat(presetsData.specificPreset[targetPresetName] || []);
```

src/utils/logger.js
```diff
@@ -13,6 +13,7 @@ const LogLevel: LogLevelType = {
   OFF: JsLogger.OFF
 };
 
+// TODO sakal check why not using playkit-js logger
```
OrenMe: In general the UI app was supposed to be playkit-js dependency free and only operate on the interface.
We tried but obviously didn't do the best-est of job, but I would like to keep it for sake of when we can define clear interface.
The real challenge here can be solved by the new bundler that can inject the logger as app wide dependency and then you avoid the duplication.
Note: you might say that the player interface has a logger, which is true and we can base upon this - open for discussion.

src/ui-manager.js
```diff
@@ -49,9 +50,17 @@ class UIManager {
     if (config.log && config.log.level && this.LogLevel[config.log.level]) {
       setLogLevel(this.LogLevel[config.log.level]);
     }
+
+    // TODO sakal check how they currently modify player log level (probably not from ui-manager)
+    if (config.log && player && player.setLogLevel && player.LogLevel) {
+      player.setLogLevel(player.LogLevel[config.log.level]);
```
OrenMe: from main API on kaltura-player

src/player-gui.js
```diff
@@ -61,7 +67,17 @@ class PlayerGUI extends Component {
     let uiToRender;
     if (this.props.uis.length > 0) {
       uiToRender = this.getMatchedUI(props.uis, props.state);
-      return uiToRender ? uiToRender.template(props) : this.props.uis[0].template(props);
+      // TODO sakal I think the fallback here is problematic and will lead to implementation issues - consider undefined instead
+      // TODO sakal  I removed the invokation of the component and left it to preact otherwise displayname are not relevant
+      const template = uiToRender ? uiToRender.template : this.props.uis[0].template;
+      const presetComponent = h(template, props);
```
OrenMe: isn't calling `h` redundant here? is this related to the second comment? if so can;t we tell it not to mangle it? or use something else? maybe it doesn't matter if it doesn't have performance implications, does it?

regarding the fallback, never has an issue with it so can't say why it matters in hindsight.
The main issue will be for custom implementations, but keep in mind not well defined states might get stuck if you don't offer default.

src/utils/logger.js
esakal: ok, will use this logger in my plugins

src/player-gui.js
esakal: calling `h` is required to create element with props.

regarding the fallback : changing the fallback to pick the last item to fix a bug

src/components/container/container-provider.js
```diff
@@ -0,0 +1,115 @@
...
+      if (!component.render || !component.container) {
+        logger.warn(
+          `preset with label '${component.label ||
+            ''}' configuration is invalid, missing required configuration (did you remember to set 'container' and 'render'?)`
+        );
+        return;
+      }
+
+      if (!component.presets) {
+        allPresets.push(component);
```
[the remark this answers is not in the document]
esakal: allow setting to all presets

src/player-gui.js
```diff
@@ -39,6 +44,7 @@ class PlayerGUI extends Component {
    * @memberof PlayerGUI
    */
   getMatchedUI(uis: Array<any>, state: Object): any {
+    // TODO sakal did you consider performance as the conditions are dynamic? why not support it explicitly
```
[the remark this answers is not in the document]
esakal: will be reconsidered later

flow-typed/types/ui-options.js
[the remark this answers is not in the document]
esakal: done

src/components/shell/shell.js
```diff
@@ -213,13 +213,12 @@ class Shell extends BaseComponent {
    * @memberof Shell
    */
   _onWindowResize(): void {
-    const playerContainer = document.getElementById(this.props.targetId);
+    // todo sakal use an exposed api
+    // todo sakal Oren - moved the `document.body` updateDocumentWidth to relevant listener
+    const playerContainer = document.getElementById(this.player._localPlayer._playerId);
```
OrenMe: don't access internals, not safe

src/ui-manager.js
```diff
@@ -62,8 +149,30 @@ class UIManager {
     this.setConfig(config);
     this._setLocaleTranslations(config);
     setEnv(this.player.env);
+
+    this._resizeWatcher = new playkitUtils.ResizeWatcher();
+    this._resizeWatcher.init(document.getElementById(config.targetId));
+    this._resizeWatcher.addEventListener(CustomEventType.RESIZE, this._onPlayerWrapperResize);
```
OrenMe: this is thrown from player, what is the use case to hold two instances? just listen to resize from player

esakal:
duplicate of #399

PR #400, opened 2019-08-06
fix(FEC-9281): ignore shortcut handling if focused on editable node

esakal:
ignore shortcut handling if focused on editable node

Solves FEC-9281


OrenMe:
I would like to refactor the keyboard shortcuts implementation and possibly make this as a HoC provider style that will allow each component to register it's own shortcut in the component did mount.
In this case I'm still debating if the responsibility to prevent propagation will be of the component or the provider.
Can this be done locally in the meanwhile until we refactor the support?

esakal:
@OrenMe  I fixed the flow typecheck issue. It appears that flow declaration is partial for PReact keyboard event so checking for the type solve it. Please review and if possible merge to next version.

esakal:
@OrenMe what about this PR? when are you planning to merge it? Note that this is essential support for us and the suggested code change in this PR is minimal and will not affect any design you would decide upon later

esakal:
@OrenMe done

PR #402, opened 2019-08-15
feat(FEC-9174): customize logo on bottom bar

Yuvalke:

add customize logo from ui config added for small screen and up

### CheckLists

- [ ] Docs have been updated


src/components/logo/logo.js
```diff
@@ -0,0 +1,77 @@
...
+  static displayName = 'logo';
+  /**
+   * should render component
+   * @param {*} props - component props
+   * @returns {boolean} - whether to render the component
+   * @static
+   */
+  static shouldRender(props: any): boolean {
+    const componentConfig = props.config.components[this.displayName];
+    return !(Object.keys(componentConfig).length === 0 && componentConfig.constructor === Object);
```
OrenMe: what is this check?

Yuvalke: this is for case that we didn't config this ui element so we didn't want to render the element.
I took it actually from watermark and I saw it in other configurable components

OrenMe: yes, I saw the reference but couldn't make sense of the actual check itself. What does it check for and why like this? you can check with @yairans who created it for the other component.

PR #403, opened 2019-08-15
fix(FEC-9079): USER_SEEKED event triggering twice

Yuvalke:

ACT: event dispatch twice on touchend and click, click & touchend fired and doing the same thing for touch devices.
SOLUTION: handle it only in touchend and fix the getTime for touch devices.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/seekbar/seekbar.js
+++ b/src/components/seekbar/seekbar.js
@@ -36,8 +36,6 @@ class SeekBarControl extends Component {
   _seekBarElement: HTMLElement;
   _framePreviewElement: HTMLElement;
   _timeBubbleElement: HTMLElement;
-  _movex: number;
-
   /**
    * Creates an instance of SeekBarControl.
    * @memberof SeekBarControl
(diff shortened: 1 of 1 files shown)


src/components/seekbar/seekbar.js
```diff
@@ -99,28 +97,6 @@ class SeekBarControl extends Component {
...
-   * @param {Event} e - onClick event
-   * @returns {void}
-   * @memberof SeekBarControl
-   */
-  onTap(e: Event): void {
-    if (!this.props.isMobile) {
-      return;
-    }
-    const oldTime = this.props.player.currentTime;
-    const newTime = this.getTime(e);
```
OrenMe: won;t this kill the tap action? we support both dragging and tapping the seekbar. Did you check?

Yuvalke: Yes

OrenMe: Yes on which one of the questions? :-)

Yuvalke: Lol, of course I checked. It's support dragging, the problem was that on click and ontouchend did same thing..

PR #404, opened 2019-08-20
fix(FEC-8797): keyboard doesn't work in fullscreen

Yuvalke:

on enter/exit fullscreen, our container should get focus for keyboard events, it won't always be the fullscreen element itself because we want the keyboard events which bound to our container. 

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/fullscreen/fullscreen.js
+++ b/src/components/fullscreen/fullscreen.js
@@ -42,7 +42,7 @@ class FullscreenControl extends BaseComponent {
    */
   toggleFullscreen(): void {
     this.logger.debug(`Toggle fullscreen`);
-    const playerContainer: HTMLElement | null = document.getElementById(this.props.config.targetId);
+    const playerContainer: HTMLElement | null = document.getElementById(this.props.targetId);
     this.props.fullscreen ? this.player.exitFullscreen() : this.player.enterFullscreen();
     if (playerContainer) playerContainer.focus();
     this.notifyClick();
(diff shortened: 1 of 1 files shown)


src/components/fullscreen/fullscreen.js
```diff
@@ -41,7 +42,9 @@ class FullscreenControl extends BaseComponent {
    */
   toggleFullscreen(): void {
     this.logger.debug(`Toggle fullscreen`);
+    const playerContainer: HTMLElement | null = document.getElementById(this.props.config.targetId);
     this.props.fullscreen ? this.player.exitFullscreen() : this.player.enterFullscreen();
+    if (playerContainer) playerContainer.focus();
```
OrenMe: no inline ifs without curly brackets

src/components/fullscreen/fullscreen.js
```diff
@@ -12,7 +12,8 @@ import {default as Icon, IconType} from '../icon';
  * @returns {Object} - mapped state to this component
  */
 const mapStateToProps = state => ({
-  fullscreen: state.engine.fullscreen
+  fullscreen: state.engine.fullscreen,
+  config: state.config
```
OrenMe: attaching state to props means also that component will update on each state change.
As you only want the targetId then target only that attribute
```suggestion
  targetId: state.config.targetId
```

src/components/fullscreen/fullscreen.js
```diff
@@ -41,7 +42,9 @@ class FullscreenControl extends BaseComponent {
    */
   toggleFullscreen(): void {
     this.logger.debug(`Toggle fullscreen`);
+    const playerContainer: HTMLElement | null = document.getElementById(this.props.config.targetId);
```
OrenMe:
```suggestion
    const playerContainer: HTMLElement | null = document.getElementById(this.props.targetId);
```

PR #415, opened 2019-09-26
fix(FEC-9375): After unmute volume icon, the volume get unmuted and then returning to unmute state

Yuvalke:

on every touch device it happened.
**issue** - `click` on volume happened after `touchend` on shell - on non-touch device it'll be `click` on volume component and then `click` on shell - the correct behaviour.
**solution** - add `touchend` event on volume button which will `stopImmediatePropagation` and shell won't get this event. as work in desktop, only `click` event will work 

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/volume/volume.js
+++ b/src/components/volume/volume.js
@@ -183,11 +183,11 @@ class Volume extends Component {
   /**
    * on volume control button click, toggle mute in player and store state
    *
-   * @method onVolumeControlButtonClick
+   * @method VolumeButtonClicked
    * @returns {void}
    * @memberof Volume
    */
-  onVolumeControlButtonClick(): void {
+  VolumeButtonClicked(): void {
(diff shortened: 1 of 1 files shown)


src/components/volume/volume.js
```diff
@@ -183,11 +183,11 @@ class Volume extends Component {
   /**
    * on volume control button click, toggle mute in player and store state
    *
-   * @method onVolumeControlButtonClick
+   * @method VolumeButtonClicked
    * @returns {void}
    * @memberof Volume
    */
-  onVolumeControlButtonClick(): void {
+  VolumeButtonClicked(): void {
```
OrenMe:
```suggestion
  onVolumeButtonClicked(): void {
```

As it is a handler keep the onX convention and make sure class methods start with lower case

PR #419, opened 2019-10-02
fix(FEC-9376):when there is fallback on ads, when clicking skip ad video is still muted

Yuvalke:

change to mouse down cause ima skip button is register to click with stop propagation.
mouse down happened which fix the issue. - occur on desktop.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/volume/volume.js
+++ b/src/components/volume/volume.js
@@ -137,7 +137,7 @@ class Volume extends Component {
    * @memberof Volume
    */
   onKeyDown(e: KeyboardEvent): void {
-    const {player} = this.state;
+    const {player} = this.props;
     /**
      * Change volume operations.
      * @param {number} newVolume - The new volume.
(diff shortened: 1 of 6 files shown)


src/components/volume/volume.js
```diff
@@ -288,7 +288,7 @@ class Volume extends Component {
           tabIndex="0"
           aria-label="Volume"
           className={style.controlButton}
-          onClick={() => this.onVolumeControlButtonClick()}
+          onMouseDown={() => this.onVolumeControlButtonMouseDown()}
```
OrenMe: Let’s change all event handlers and omit the VolumeControl part, i.e. onMouseDoen, onKeyDown

OrenMe:
@Yuvalke you mean the IMA SDK rendered button has skip propagation? Maybe better then to register to skip event?

PR #440, opened 2019-10-27
fix(FEC-9450): keyboard prevent non relevant events

Yuvalke:

Special key like alt, ctrl and cmd didn't work with keys that already mapped.
prevent only combine keys with handlers.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/keyboard/keyboard.js
+++ b/src/components/keyboard/keyboard.js
@@ -70,11 +70,14 @@ class Keyboard extends Component {
     }
     playerContainer.onkeydown = (e: KeyboardEvent) => {
       if (!this.props.shareOverlay && !this.props.playerNav && typeof this.keyboardHandlers[e.keyCode] === 'function') {
-        e.preventDefault();
         this.props.logger.debug(`KeyDown -> keyName: ${getKeyName(e.keyCode)}, shiftKey: ${e.shiftKey.toString()}`);
-        const payload = this.keyboardHandlers[e.keyCode](e);
-        if (payload) {
-          this.props.notifyClick({key: e.keyCode, ...payload});
+        const keyHandler = this.keyboardHandlers[e.keyCode](e);
+        if (keyHandler) {
+          e.preventDefault();
+          const payload = keyHandler[e.keyCode];
+          if (payload) {
+            this.props.notifyClick({key: e.keyCode, ...payload});
+          }
         }
       }
     };
@@ -112,7 +115,7 @@ class Keyboard extends Component {
         this.props.updateOverlayActionIcon(IconType.Play);
       }
       this.toggleHoverState();
-      return true;
+      return {[KeyMap.SPACE]: true};
     },
     [KeyMap.UP]: () => {
       let newVolume = (Math.round(this.props.player.volume * 100) + KEYBOARD_DEFAULT_VOLUME_JUMP) / 100;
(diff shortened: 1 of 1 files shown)


src/components/keyboard/keyboard.js
```diff
@@ -105,8 +104,8 @@ class Keyboard extends Component {
    *
    * @memberof Keyboard
    */
-  keyboardHandlers: {[key: number]: Function} = {
-    [KeyMap.SPACE]: () => {
+  keyboardHandlers: {[key: number]: (event: KeyboardEvent) => {preventDefault: boolean, payload: any}} = {
```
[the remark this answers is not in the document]
Yuvalke: done

flow-typed/types/event-handler-option.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+declare type EventHandlerOption = {
```
OrenMe: this name is extremely generic and will apply to anything which is an event, while it is a specific form for a specific handler.
```suggestion
declare type KeyboardEventResult = {
```

Yuvalke: we could use it in other cases like this, why do you want to rename it to specific existing use case?

OrenMe: cause it is specific, where else will you have prevent default and payload response?
The response object is tightly coupled with the implementation of this class

Yuvalke: we didn't use, it could be used on the future..
it's use case I think.

OrenMe: unless you see a concrete use case then it is specific.
refactor to generalize a use case later is always possible, but there is no benefit or use case here.
Let's change it to fit the part.

Yuvalke: OK

src/components/keyboard/keyboard.js
```diff
@@ -184,85 +189,99 @@ class Keyboard extends Component {
...
+      if (event.shiftKey) {
+        if (this.props.player.playbackRate !== this.props.player.defaultPlaybackRate) {
+          this.props.logger.debug(`Changing playback rate. ${this.props.player.playbackRate} => ${this.props.player.defaultPlaybackRate}`);
+          this.props.player.playbackRate = this.props.player.defaultPlaybackRate;
+          this.props.updateOverlayActionIcon(IconType.Speed);
+          return {preventDefault: true, payload: {speed: this.props.player.defaultPlaybackRate}};
+        }
+        return {preventDefault: true, payload: null};
       }
+      return {preventDefault: false, payload: null};
```
OrenMe: if `{preventDefault: false, payload: null}` repeats everywhere extract it to a const and use that.
`const unhandledKeyboardEventResult = {preventDefault: false, payload: null};`

src/components/keyboard/keyboard.js
```diff
@@ -303,6 +303,15 @@ class Keyboard extends Component {
...
     }, CONTROL_BAR_HOVER_DEFAULT_TIMEOUT);
   }
+
+  /** get unhandled value for keyboard event
+   *
+   * @returns {KeyboardEventResult} - default result
+   * @memberof Keyboard
+   */
+  get unhandledKeyboardEventResult(): KeyboardEventResult {
+    return {preventDefault: false, payload: null};
```
OrenMe: as it static const make it a `const unhandledKeyboardEventResult: KeyboardEventResult = {preventDefault: false, payload: null};`
place it in top of file and use it internally.

PR #453, opened 2019-11-12
fix(FEC-9433): timer on seek preview displayed as NaN:NaN:NaN

RoyBregman:

when dragging scrubber to 0 it fell back to e.changedTouches which was undefined.
Changed the logic to first use e.clientX if is a number (zero included)

Fixes FEC-9433

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/seekbar/seekbar.js
+++ b/src/components/seekbar/seekbar.js
@@ -343,7 +343,7 @@ class SeekBar extends Component {
    * @memberof SeekBar
    */
   getTime(e: any): number {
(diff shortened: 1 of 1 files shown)


OrenMe:
@RoyBregman what happens when you scrub all the way to the end(right side)?

RoyBregman:
> @RoyBregman what happens when you scrub all the way to the end(right side)?

It will work as today - The "bug" was only that clientx was ignored when 0 when actually 0 is okay
I rewrote it again with Yair, maybe now it is more readable

PR #454, opened 2019-11-13
fix(FEC-9493): ima - dragging to the end causes the midroll to play with replay button

yairans:

**Issue:** ima plays the midroll on `ENDED` asynchronously, so `PLAYBACK_ENDED` fired before `AD_LOADED` and `AD_BREAK_START`.
**Solution:** Hide the replay button while ad break, even `isPlaybackEnded` prop is `true`

(This is for ima vmap. The same issue with ad layout fixed by https://github.com/kaltura/playkit-js/pull/416)

Solves [FEC-9493](https://kaltura.atlassian.net/browse/FEC-9493)

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/play-pause/play-pause.js
+++ b/src/components/play-pause/play-pause.js
@@ -56,25 +56,16 @@ class PlayPause extends Component {
    */
   render(props: any): React$Element<any> | void {
     const controlButtonClass = this.props.isPlayingAdOrPlayback ? [style.controlButton, style.isPlaying].join(' ') : style.controlButton;
+    const isStartOver = props.isPlaybackEnded && !this.props.adBreak;
     return (
       <div className={[style.controlButtonContainer, style.controlPlayPause].join(' ')}>
         <Localizer>
           <button
             tabIndex="0"
-            aria-label={
-              <Text
-                id={
-                  props.isPlaybackEnded && !this.props.adBreak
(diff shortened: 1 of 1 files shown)


src/components/play-pause/play-pause.js
```diff
@@ -62,11 +62,19 @@ class PlayPause extends Component {
           <button
             tabIndex="0"
             aria-label={
-              <Text id={props.isPlaybackEnded ? 'controls.startOver' : this.props.isPlayingAdOrPlayback ? 'controls.pause' : 'controls.play'} />
+              <Text
+                id={
+                  props.isPlaybackEnded && !this.props.adBreak
```
RoyBregman: extract to const isStartOver to be used also in line 77

OrenMe:
How does this happen @yairans ? playback ended should happen only after all ads were played, so the new condition is inherently incorrect.
So it looks like we have some other issue which is not necessarily in the UI - where is it?

yairans:
@OrenMe as I wrote in the description, ima triggers the midroll on ended **asynchronously**. We don’t know in our ended handler that an ad will play soon and we trigger playback-ended

src/components/play-pause/play-pause.js
yairans: done

PR #459, opened 2019-11-28
fix(FEC-9471): slider progress bar exceeds 100%

RoyBregman:

liveedge is not calculated anymore by a threshold but received from the player
In addition - when in live edge the progress bar duration is the current time

solves FEC-9471

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/seekbar-live-playback-container/seekbar-live-playback-container.js
+++ b/src/components/seekbar-live-playback-container/seekbar-live-playback-container.js
@@ -78,7 +78,7 @@ class SeekBarLivePlaybackContainer extends Component {
         updateCurrentTime={data => this.props.updateCurrentTime(data)}
         isDvr={this.props.isDvr}
         currentTime={this.props.currentTime}
-        duration={this.props.player.isOnLiveEdge() ? this.props.currentTime : this.props.duration}
+        duration={this.props.player.isOnLiveEdge() ? this.props.currentTime : Math.max(this.props.duration, this.props.currentTime)}
         isDraggingActive={this.props.isDraggingActive}
         isMobile={this.props.isMobile}
         notifyChange={payload => this.props.notifyChange(payload)}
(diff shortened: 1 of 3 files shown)


src/components/seekbar-live-playback-container/seekbar-live-playback-container.js
```diff
@@ -78,7 +78,7 @@ class SeekBarLivePlaybackContainer extends Component {
         updateCurrentTime={data => this.props.updateCurrentTime(data)}
         isDvr={this.props.isDvr}
         currentTime={this.props.currentTime}
-        duration={this.props.duration}
+        duration={this.props.player.isOnLiveEdge() ? this.props.currentTime : Math.max(this.props.duration, this.props.currentTime)}
```
yairans:
```suggestion
        duration={this.props.player.isOnLiveEdge() ? this.props.currentTime : this.props.duration}
```
`currentTime > duration` means `isOnLiveEdge()` is `true` so the `Math.max` is unnecessary

RoyBregman: No, because the check of the isOnLiveEdge is done on seeking. But it can be that in seeking the currentTime was smaller than duration (hence isOnLiveEdge = false) but because segments are being downloaded and duration "jumps" than its possible currentTime will be greater than duration and it will exceed 100%. This is another fail safe mechanism

src/components/seekbar-live-playback-container/seekbar-live-playback-container.js
```diff
@@ -71,20 +71,33 @@ class SeekBarLivePlaybackContainer extends Component {
       <SeekBar
         playerElement={this.props.playerContainer}
         showTimeBubble={this.props.showTimeBubble}
-        changeCurrentTime={time => (this.props.player.currentTime = time)}
+        changeCurrentTime={time => {
+          if (!this.props.player.isOnLiveEdge() || time < this.duration) {
```
Yuvalke: if !(this.props.player.isOnLiveEdge() && time === this.duration)

RoyBregman: fixed

PR #460, opened 2019-11-28
fix(FEC-9505): Add i18n support for custom captions

RoyBregman:

replaced span with text tag and id of new localization

solves FEC-9505

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/translations/en.json
+++ b/src/translations/en.json
@@ -59,6 +59,7 @@
   "cvaa": {
     "title": "Advanced captions settings",
     "sample_caption_tag": "Sample",
+    "sample_custom_caption_tag": "Custom captions",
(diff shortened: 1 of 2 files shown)


src/translations/en.json
```diff
@@ -59,6 +59,7 @@
   "cvaa": {
     "title": "Advanced captions settings",
     "sample_caption_tag": "Sample",
+    "sample_custom_caption_tag": "Custom captions",
```
Yuvalke: I think you need to keep the format of - "first_second": "First second", I mean to remove the sample

RoyBregman: The reason i left the "sample" was that this is also a sample that shows how the captions look like.

PR #464, opened 2019-12-05
feat(FEC-9481): distributed keyboard handler to decorator

Yuvalke:

create keyboard decorator event to implement in each component which use keyboard event on player area.

Solves: FEC-9481

### CheckLists

- [ ] Docs have been updated


src/components/keyboard/with-keyboard-event.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+const withKeyboardEvent = (WrappedComponent: Component) => {
+  return class KeyboardComponent extends Component {
+    /**
+     * add keyboard handler
+     * @param {number} keyCode - Optional payload.
+     * @param {Function} callback - Optional payload.
+     * @returns {void}
+     *
+     * @memberof EventDispatcherComponent
```
yairans: fix

src/components/keyboard/keyboard-event-provider.js
```diff
@@ -0,0 +1,121 @@
...
+  _createKeyCode(key: KeyboardKey): number {
+    const altKey = key.altKey ? 1 : 0;
+    const ctrlKey = key.ctrlKey ? 1 : 0;
+    const metaKey = key.metaKey ? 1 : 0;
+    const shiftKey = key.shiftKey ? 1 : 0;
+    return parseInt('' + key.code + altKey + ctrlKey + metaKey + shiftKey);
+  }
+  /**
+   * create context player
+   * @returns {void}
```
yairans: fix

src/components/keyboard/with-keyboard-event.js
```diff
@@ -0,0 +1,35 @@
...
+     */
+    addKeyboardHandler(keyCode: KeyboardKey, callback: Function): void {
+      this.context.addKeyboardHandler(keyCode, callback);
+    }
+
+    /**
+     * render component
+     *
+     * @returns {React$Element} - component element
+     * @memberof LoggerComponent
```
yairans: fix

src/components/language/language.js
```diff
@@ -76,6 +79,38 @@ class Language extends Component {
    */
   componentDidMount() {
     this.props.eventManager.listen(document, 'click', e => this.handleClickOutside(e));
+    this.props.addKeyboardHandler({code: KeyMap.C}, event => {
+      event.preventDefault();
```
yairans: import this handler from `keyboard.js`. we can't maintain code like this twice.
same for all new handlers in this change

src/components/keyboard/keyboard-event-provider.js
```diff
@@ -0,0 +1,121 @@
...
+      this._keyboardListeners[keyCombine](event);
+    }
+  }
+  /**
+   * check if keyboard handler should be used
+   * @returns {boolean} - handler should be used
+   * @private
+   */
+  _shouldHandledKeyboardEvent(): boolean {
+    return this.props.isKeyboardEnable && !this.props.shareOverlay && !this.props.playerNav;
```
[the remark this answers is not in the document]
Yuvalke: No, if user will change the `isKeyboardEnable` to `false` and then share will change it to `false` as well and then `true` the final state of `isKeyboardEnable` will be `true` and that's not what the user wants.

src/components/language/language.js
Yuvalke: we'll not.
`keyboard.js` is deprecated if someone wants to use it he'll need to write the logic by himself I don't think we need to maintain this code anymore.

src/components/language/language.js
```diff
@@ -55,9 +57,36 @@ const COMPONENT_NAME = 'Language';
 class Language extends Component {
   state: Object;
   _controlLanguageElement: any;
+  _lastActiveTextLanguage: string = '';
   // ie11 fix (FEC-7312) - don't remove
   _portal: any;
-
+  _keyboardEventHandlers: Array<KeyboardEventHandlers> = [
+    {
+      eventType: 'keydown',
```
OrenMe: each event should have 
{
  eventType?,
  key,
  handler
}

cause each key binding can have different eventType potentially
Also, eventType should be optional cause keyDown is the default

Yuvalke: I grouped by eventType to make it clear
[{eventType:keydown, 
handlers:[]}]

Yuvalke: Handler -> {key:keyboardKey, action:function}

PR #466, opened 2019-12-10
fix(FEC-9555): in mobile devices portrait mode when countdown shows the player jumps to the left

RoyBregman:

commenting out the auto focus
deprecating FEC-9479 for now
solves FEC-9555

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/playlist-countdown/playlist-countdown.js
+++ b/src/components/playlist-countdown/playlist-countdown.js
@@ -152,7 +152,7 @@ class PlaylistCountdown extends Component {
     }
 
     if (!prevState.shown && this.state.shown && this.focusElement) {
-      this.focusElement.focus();
+      this.focusElement.focus({preventScroll: true});
     }
 
     if (this.isShown !== this.state.shown) this.setState({shown: this.isShown});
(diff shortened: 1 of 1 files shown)


OrenMe:
@yairans what are the possible options?
For preventscroll there is a supposed polyfill https://github.com/calvellido/focus-options-polyfill/blob/master/index.js
As previously implemented we can wait for transition to end.
Do you have other solutions?

yairans:
@OrenMe I would check first maybe `preventScroll` indeed supported cross browsers.
also, yesterday @RoyBregman and I saw that it can be fixed by css

RoyBregman:
checking if the browser supports preventScroll doesn't seem to be possible as it is a parameter passed to the focus. There is no pre-check that can be done. Only maybe a post check to see that there was no scrolling.

CSS changes the way we talked about yesterday are also not ideal as they ruin the css effect.

I am working on another approach. will update in the following hour or two

OrenMe:
@RoyBregman see the polyfill I linked, seems it is supported.
Let’s analyze possibilities and choose the one that does the job with least risk.

PR #467, opened 2019-12-12
feat(FEC-9479): focus on countdown without scrolling issue

RoyBregman:
Separated the slide from right as an animation and added a listener to animationend to autofocus and to make the component accessible.
In addition, components "next" will be updated only when shown to avoid seeing the next video of playlist when fading out.

solves FEC-9479

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/playlist-countdown/_playlist-countdown.scss
+++ b/src/components/playlist-countdown/_playlist-countdown.scss
@@ -1,5 +1,5 @@
 .player {
-  .playlist-countdown-yOffset {
+  .playlistCountdownYOffset {
     transform: translateY(0px);
     transition: transform ease-in 100ms;
     width: 100%;
@@ -115,7 +115,7 @@
     &.state-paused,
     &.state-idle,
     &.hover {
-      .playlist-countdown-yOffset {
+      .playlistCountdownYOffset {
         transform: translateY(-50px);
       }
     }
--- a/src/components/playlist-countdown/_playlist-countdown.scss
+++ b/src/components/playlist-countdown/_playlist-countdown.scss
@@ -1,5 +1,10 @@
 .player {
-
+  .playlistCountdownYOffset {
+    transform: translateY(0px);
+    transition: transform ease-in 100ms;
+    width: 100%;
+    height: 100%;
+  }
   .playlist-countdown {
     font-family: $font-family;
     height: 72px;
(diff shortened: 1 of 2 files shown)


src/components/playlist-countdown/_playlist-countdown.scss
```diff
@@ -1,18 +1,24 @@
 .player {
-
+  .playlist-countdown-yOffset {
```
yairans: move it into `.playlist-countdown` class to keep the hierarchy

src/components/playlist-countdown/playlist-countdown.js
```diff
@@ -152,8 +152,7 @@ class PlaylistCountdown extends Component {
     }
 
     if (!prevState.shown && this.state.shown && this.focusElement) {
-      // deprecated for now due to scrolling bug in portrait android
-      // this.focusElement.focus({preventScroll: true});
+      this.focusElement.focus({preventScroll: true});
```
yairans:
```suggestion
      this.focusElement.focus();
```

src/components/playlist-countdown/_playlist-countdown.scss
```diff
@@ -1,5 +1,15 @@
+@keyframes slideIn {
+  0% {
```
yairans: 0% - 50% is for delay? 
if so, use the `animation-delay` property:
`animation: slideIn 500ms 500ms` should work

src/components/playlist-countdown/_playlist-countdown.scss
```diff
@@ -8,11 +18,15 @@
...
+    transition: transform ease-in 100ms, opacity 400ms, bottom 300ms;
     cursor: pointer;
 
+    &.slideIn{
+      animation: slideIn 1000ms
+    }
+
     &.hidden {
-      right: -160px;
+      transition: opacity 50ms;
```
yairans: This causes a bad animation once moving from show to hidden 

Kaltura-Player-V3-_1_

src/components/playlist-countdown/_playlist-countdown.scss
RoyBregman: i want it to start at -160px. if i use delay instead then it will be at right 0px while fading in

yairans: so return `right: -160px` in `.hidden`

PR #468, opened 2019-12-18
feat(FEC-7804): implement tooltips

RoyBregman:

Implemented tooltip support - 

- contains 8 types of tooltips (default top)
- boundary check and fallback to other types in case of invalid type
- multiline and max-width support (default 240px. none will make it one line)
- next and prev tooltips will not be shown in case the poster preview is shown
- added debounce util to limit a burst of function calls

Solves (FEC-7804)

### CheckLists

- [ ] Docs have been updated


src/components/cast/cast.js
```diff
@@ -58,17 +61,21 @@ class Cast extends Component {
...
-          onKeyDown={e => {
-            if (e.keyCode === KeyMap.ENTER) {
-              this.props.updateBackdropVisibility(true);
-              this.props.player.startCasting().catch(() => this.props.updateBackdropVisibility(false));
-            }
-          }}>
-          <google-cast-launcher className={style.castButton} tabIndex="0" />
-        </div>
+        <Tooltip label={this.props.castText}>
+          <div
```
Yuvalke: why doesn't it button?

src/components/play-pause/play-pause.js
```diff
@@ -57,14 +63,11 @@ class PlayPause extends Component {
   render(props: any): React$Element<any> | void {
     const controlButtonClass = this.props.isPlayingAdOrPlayback ? [style.controlButton, style.isPlaying].join(' ') : style.controlButton;
     const isStartOver = props.isPlaybackEnded && !this.props.adBreak;
+    const localization = isStartOver ? this.props.startOverText : this.props.isPlayingAdOrPlayback ? this.props.pauseText : this.props.playText;
```
Yuvalke: Split it to two rows, we don't use inline if in inline if. and more valuable name.
```
const buttonStateText = this.props.isPlayingAdOrPlayback ? this.props.pauseText : this.props.playText;
const buttonLabelText = isStartOver ? this.props.startOverText : buttonText;
```

src/components/playlist-button/playlist-button.js
```diff
@@ -46,44 +53,57 @@ class PlaylistButton extends Component {
...
-                <Icon type={IconType.Prev} />
-              </div>
-            ) : (
-              <div>
-                <Icon type={IconType.Next} />
-              </div>
-            )}
-          </button>
-        </Localizer>
+        {showPreview ? (
```
Yuvalke: don't you need tooltip here as well?

src/components/settings/settings.js
```diff
@@ -211,13 +212,15 @@ class Settings extends Component {
...
     return (
       <div ref={c => (this._controlSettingsElement = c)} className={[style.controlButtonContainer, style.controlSettings].join(' ')}>
-        <button
-          tabIndex="0"
-          aria-label={props.buttonAriaLabel}
-          className={this.state.smartContainerOpen ? [style.controlButton, style.active].join(' ') : style.controlButton}
-          onClick={() => this.onControlButtonClick()}>
-          <Icon type={IconType.Settings} />
-        </button>
+        <Tooltip label={this.props.buttonLabel}>
```
Yuvalke: make compatible with line 218

Yuvalke: I mean with `this` or without

src/components/tooltip/tooltip.js
```diff
@@ -10,6 +35,159 @@ import {h, Component} from 'preact';
...
+    this.setState({showTooltip: false, marginTop: 0, marginLeft: 0});
+  }
+
+  /**
+   * gets the tooltip position based on component prop
+   *
+   * @returns {void}
+   * @memberof Tooltip
+   */
+  setMargins() {
```
Yuvalke: why don't you create css class and just set the correct class? I think it's too much style work on js.

src/components/tooltip/tooltip.js
```diff
@@ -10,6 +35,159 @@ import {h, Component} from 'preact';
...
+
+  /**
+   * checks if the current tooltip type is within the player boundaries
+   * @memberof Tooltip
+   * @returns {string} is in boundaries
+   */
+  isToolTipInBoundaries(): boolean {
+    const tooltipBoundingRect = this.textElement.getBoundingClientRect();
+    const playerContainerRect = this.props.playerClientRect;
+    const MARGIN = 10;
```
Yuvalke: You have here constant like margin, that you are using it on css as well.
Could it be in css?

src/components/tooltip/_tooltip.scss
```diff
@@ -1,31 +1,122 @@
...
+    &:after {
+      content: '';
+      position: absolute;
+      border-width: 5px;
+      border-style: solid;
+      border-radius: 3px;
+      color: $tooltip-bg-color;
+      z-index: -1;
+    }
+    &.tooltip-top {
```
Yuvalke: the constant here should be same for left and right, bottom and top right?
Could it be in constant ?

src/components/tooltip/tooltip.js
OrenMe: magic numbers should be in the top of the file as CAPITAL snake case

src/components/cast/cast.js
RoyBregman: This is old code - when working on the accessibility i tried changing it to button but the css was ruined so i rolled it back

src/components/play-pause/play-pause.js
RoyBregman: i agree that inline if inside another inline if is not readable - but anyway the result is just one const not two. i will change it and see if you agree also

src/components/playlist-button/playlist-button.js
RoyBregman: No. when there is a preview it is ugly and unnecessary to have also a tooltip

src/components/settings/settings.js
RoyBregman: fixed

src/components/tooltip/_tooltip.scss
RoyBregman: agree, fixed

src/components/tooltip/tooltip.js
RoyBregman: Yep :) done

src/components/tooltip/tooltip.js
RoyBregman: This is not the same margin as in the css - this represents the player margins (So the tooltip will reach until the last pixels). I renamed it to PLAYER_MARGIN and moved to the top like Oren mentioned.

src/components/tooltip/tooltip.js
```diff
@@ -10,6 +36,116 @@ import {h, Component} from 'preact';
...
+
+  /**
+   * default component props
+   * @type {Object}
+   * @memberof Tooltip
+   */
+  static defaultProps = {
+    type: ToolTipType.Top,
+    isMultiline: true,
+    maxWidth: '240px'
```
Yuvalke: create const

src/components/tooltip/tooltip.js
```diff
@@ -18,15 +154,23 @@ class Tooltip extends Component {
...
-    let className = [style.tooltip];
-    if (props.out) className.push(style.out);
+    const className = [style.tooltipLabel, style[`tooltip-${this.state.type}`]];
+    this.state.showTooltip ? className.push(style.show) : className.push(style.hide);
 
     return (
-      <div className={className.join(' ')} style={props.left ? {left: props.left} : ''}>
+      <div className={style.tooltip} onMouseOver={() => this.onMouseOver()} onMouseLeave={() => this.onMouseLeave()}>
+        <span
+          style={{maxWidth: props.isMultiline ? props.maxWidth : 'none'}}
```
Yuvalke: when this state change?

src/components/tooltip/tooltip.js
```diff
@@ -10,6 +36,116 @@ import {h, Component} from 'preact';
...
+  }
+
+  /**
+   * checks if after the render the tooltip is within boundaries of the player
+   * if not it will try to set a new type which will be checked after the next render
+   * @memberof Tooltip
+   * @returns {void}
+   */
+  componentDidUpdate(): void {
+    if (this.state.showTooltip && !this.isToolTipInBoundaries()) {
```
Yuvalke: Don't you can test the all possible tooltips and select the one which can be used by order?
Here you make array of non valid type but the playerClientRect can change and they will be relevant again

src/components/tooltip/_tooltip.scss
```diff
@@ -1,31 +1,129 @@
...
+      bottom: $tooltip-margin;
+      left: 50%;
+      transform: translate(-50%);
+      &:after {
+        bottom: -$tooltip-arrow-height;
+        left: 50%;
+        transform: translate(-50%) rotate(45deg) skew(5deg, 5deg);
+      }
+    }
+    &.tooltip-top-left {
```
yairans: in generally `.tooltip-top-left` should be a combination of `.tooltip-top` and `.tooltip-left` as we do in the watermark component. does it applicable here?
any way, `.tooltip-top-left` and `.tooltip-top-right` should use `.tooltip-top` rules. only overide the different rule. 
try to reuse the rules as much as possible

src/components/tooltip/tooltip.js
```diff
@@ -1,6 +1,32 @@
...
+/**
+ * mapping state to props
+ * @param {*} state - redux store state
+ * @returns {Object} - mapped state to this component
+ */
+const mapStateToProps = state => ({
+  playerClientRect: state.shell.playerClientRect
+});
+
+export const TOOLTIP_SHOW_TIMEOUT: number = 250;
```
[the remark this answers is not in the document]
RoyBregman: there isn't. i will remove the export

src/components/tooltip/tooltip.js
RoyBregman: this is kind of constant the way it is now in the defaultProps. it is single place to update the defaults. i think it is clear the way it is now

src/components/tooltip/tooltip.js
RoyBregman: is this still relevant?

src/components/tooltip/tooltip.js
RoyBregman: fixed

src/components/tooltip/tooltip.js
```diff
@@ -10,6 +36,123 @@ import {h, Component} from 'preact';
...
+   * brings another tooltip type which hasn't been marked as invalid
+   * @memberof Tooltip
+   * @returns {string} tooltip type
+   */
+  getAlternateType(): string {
+    let nextAlternativeType: string = null;
+    Object.keys(ToolTipType).forEach((key, index) => {
+      if (!nextAlternativeType && index > this.lastAlternativeTypeIndex && ToolTipType[key] !== this.state.type) {
+        this.lastAlternativeTypeIndex = index;
+        nextAlternativeType = ToolTipType[key];
```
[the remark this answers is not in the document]
RoyBregman: done

src/components/tooltip/tooltip.js
```diff
@@ -10,6 +37,117 @@ import {h, Component} from 'preact';
...
+   * brings another tooltip type which hasn't been marked as invalid
+   * @memberof Tooltip
+   * @returns {string} tooltip type
+   */
+  getAlternateType(): string {
+    const alternatives = Object.values(ToolTipType).filter(item => {
+      return item != this.props.type;
+    });
+    this.lastAlternativeTypeIndex = Math.min(this.lastAlternativeTypeIndex + 1, alternatives.length - 1);
+    return ((alternatives[this.lastAlternativeTypeIndex]: any): string);
```
yairans:
```
return Object.values(ToolTipType).find((item, index) => {
    if (index > this.lastAlternativeTypeIndex && item != this.props.type) {
        this.lastAlternativeTypeIndex = index;
        return true;
    } else {
        return false;
    }
});
```

src/components/shell/shell.js
```diff
@@ -214,6 +214,9 @@ class Shell extends Component {
     const {isIPadOS, isTablet, isMobile} = player.env;
     this.props.updateIsMobile(isIPadOS || isTablet || isMobile || forceTouchUI);
     this._onWindowResize();
+    window.addEventListener('resize', () => {
```
yairans:
```suggestion
    this.props.eventManager.listen(window, player.Event.RESIZE, () => this._onWindowResize());
```

src/components/tooltip/tooltip.js
```diff
@@ -10,6 +37,123 @@ import {h, Component} from 'preact';
...
+   * checks if after the render the tooltip is within boundaries of the player
+   * if not it will try to set a new type which will be checked after the next render
+   * @param {Object} prevProps - previous component props
+   * @memberof Tooltip
+   * @returns {void}
+   */
+  componentDidUpdate(prevProps: Object): void {
+    if (this.props.playerClientRect !== prevProps.playerClientRect) {
+      this.lastAlternativeTypeIndex = -1;
+      this.setState({type: this.props.type});
```
yairans:
```suggestion
      this.setState({valid: false, type: this.props.type});
```
as you said yesterday. no?

src/components/shell/shell.js
RoyBregman: fixed, thanks

src/components/tooltip/tooltip.js
RoyBregman: yes, you actually listen to what i say :))) thanks!

src/components/shell/shell.js
yairans: also see https://github.com/kaltura/playkit-js-ui/pull/367 and verify this issue with your change

OrenMe: use the event manager for this as well

src/components/shell/shell.js
```diff
@@ -214,7 +214,10 @@ class Shell extends Component {
     const {isIPadOS, isTablet, isMobile} = player.env;
     this.props.updateIsMobile(isIPadOS || isTablet || isMobile || forceTouchUI);
     this._onWindowResize();
-    this.props.eventManager.listen(player, player.Event.RESIZE, () => this._onWindowResize());
+    this.props.eventManager.listen(window, 'resize', () => {
```
Yuvalke: why did you change it to window?

src/components/tooltip/tooltip.js
```diff
@@ -10,6 +37,123 @@ import {h, Component} from 'preact';
...
+    this._hoverTimeout = null;
+  }
+
+  /**
+   * brings another tooltip type which hasn't been marked as invalid
+   * @memberof Tooltip
+   * @returns {string} tooltip type
+   */
+  getAlternateType(): ?string {
+    return ((Object.values(ToolTipType).find((item, index) => {
```
Yuvalke: `const ToolTipTypeValues = Object.values(ToolTipType)
ToolTipTypeValues.length > this.lastAlternativeTypeIndex ? this.lastAlternativeTypeIndex++ : this.lastAlternativeTypeIndex
return ToolTipTypeValues[this.lastAlternativeTypeIndex]`

src/components/tooltip/tooltip.js
```diff
@@ -10,6 +37,123 @@ import {h, Component} from 'preact';
...
+    maxWidth: '240px'
+  };
+
+  /**
+   * @constructor
+   * @param {*} props props
+   * @memberof Tooltip
+   */
+  constructor(props: Object) {
+    super(props);
```
OrenMe: no need for constructor if there's no implementation in it

src/components/tooltip/_tooltip.scss
RoyBregman: not possible as we discussed

src/components/tooltip/_tooltip.scss
```diff
@@ -1,31 +1,129 @@
...
-    height: 10px;
-    background-color: #fff;
-    border-radius: 3px;
-    transform: rotate(45deg) translate3d(0%, 50%, 0);
-    left: 50%;
-    bottom: -1px;
+    line-height: normal;
+    font-weight: bold;
+    font-size: 13px;
+    z-index: 2;
```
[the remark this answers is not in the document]
RoyBregman: done

src/components/tooltip/tooltip.js
```diff
@@ -10,6 +36,123 @@ import {h, Component} from 'preact';
...
+    this._hoverTimeout = null;
+  }
+
+  /**
+   * brings another tooltip type which hasn't been marked as invalid
+   * @memberof Tooltip
+   * @returns {string} tooltip type
+   */
+  getAlternateType(): string {
+    let nextAlternativeType: string = null;
```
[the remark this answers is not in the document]
RoyBregman: outdated

src/components/tooltip/tooltip.js
RoyBregman: removed

src/components/tooltip/tooltip.js
RoyBregman: i am leaving it as it is

src/components/shell/shell.js
RoyBregman: done

src/components/tooltip/tooltip.js
RoyBregman: fixed

PR #485, opened 2020-02-13
feat(FEC-9465): internationalization (i18n) - player localization

yairans:

* Move `translations` folder to the root to be available to kaltura-player (https://github.com/kaltura/kaltura-player-js/pull/304)
* Add French translation 

Solves FEC-9465

### CheckLists

- [ ] Docs have been updated


src/ui-manager.js
```diff
@@ -39,7 +39,7 @@ class UIManager {
   store: any;
   container: ?HTMLElement;
   root: React$Component<any, any, any>;
-  _translations: {[langKey: string]: Object} = {en: en_translations};
+  _translations: {[langKey: string]: Object} = {en: en_translations['en']};
```
OrenMe: there is no need for the extra nesting IMO, the file is already called by the key name

yairans: else the bundle builder should add the key name to the generated i18n json.

PR #497, opened 2020-04-19
fix(FEC-9980): custom preset controls are continuously blinking on hovering the bottom bar

yairans:

**Problem**: Once the hovering timeout is done the shell is updating the store with `playerHover:false` which causes a renderer, but at this time the mouse is still hovering the shell so in `onMouseOver` the shell is updating the store with `playerHover:true`, and so on for ever...
(This reproduce in the preact version 8.x (current) only when custom preset is set, but in preactX even with our default presets)

**Solution**: remove the state update from `onMouseOver` as it's redundant since `onMouseMove`

Solves: FEC-9980

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/shell/shell.js
+++ b/src/components/shell/shell.js
@@ -101,9 +101,6 @@ class Shell extends Component {
       this.setState({nav: false});
       this.props.updatePlayerNavState(false);
     }
-    if (!this.props.bottomBarHoverActive) {
-      this._updatePlayerHoverState();
-    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/components/shell/shell.js
```diff
@@ -412,7 +412,6 @@ class Shell extends Component {
         className={playerClasses}
         onTouchEnd={e => this.onTouchEnd(e)}
         onMouseUp={() => this.onMouseUp()}
-        onMouseOver={() => this.onMouseOver()}
```
Yuvalke, line 415: delete the handler on `onMouseOver`

src/components/shell/shell.js
```diff
@@ -412,7 +412,6 @@ class Shell extends Component {
...
         onMouseMove={() => this.onMouseMove()}
```
Yuvalke, line 415: mouse move doesn't handle nav state, it means on keyboard nav it'll stay open -> it handled on `onMouseOver `

PR #512, opened 2020-06-01
fix(FEC-10144): player causes form submit on button click

shalommeoded:

When the player is nested within a form when clicking any button that is in the control bar we get redirected to any form referral.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -1,0 +1,27 @@
+//@flow
+import {h} from 'preact';
+import {Component} from 'preact';
+
+const COMPONENT_NAME = 'Button';
+
+/**
+ * Button component
+ *
+ * @class Button
+ * @example <Button/>
+ * @extends {Component}
+ */
+class Button extends Component {
+  /**
+   * render component
+   * @param {*} props - component props
+   * @returns {?React$Element} - component element
+   * @memberof Button
+   */
+  render(props: any): ?React$Element<any> {
+    return <button type={'button'} {...props} />;
+  }
+}
+
+Button.displayName = COMPONENT_NAME;
+export {Button};
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -8,7 +8,6 @@ const COMPONENT_NAME = 'Button';
  *
  * @const Button
  * @example <Button/>
- * @extends {Component}
  */
 const Button = forwardRef((props, ref) => <button type="button" ref={ref} {...props} />);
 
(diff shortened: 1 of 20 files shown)


src/components/button/button.js
```diff
@@ -0,0 +1,27 @@
...
+const COMPONENT_NAME = 'Button';
+
+/**
+ * Button component
+ *
+ * @class Button
+ * @example <Button/>
+ * @extends {Component}
+ */
+class Button extends Component {
```
OrenMe, line 14: as we don't need any lifecycle management we can use function component

OrenMe:
@shalommeoded what do you mean by collide with the `withAnimation` HoC?

src/components/forward/forward.js
```diff
@@ -66,14 +67,15 @@ class Forward extends Component {
     return (
       <div className={[style.controlButtonContainer, style.noIdleControl].join(' ')}>
         <Tooltip label={this.props.forwardText}>
-          <button
+          <Button
+            type={'button'}
```
yairans, line 71: is needed? 
already got it in Button no?

src/components/rewind/rewind.js
```diff
@@ -65,14 +66,15 @@ class Rewind extends Component {
     return (
       <div className={[style.controlButtonContainer, style.noIdleControl].join(' ')}>
         <Tooltip label={this.props.rewindText}>
-          <button
+          <Button
+            type={'button'}
```
yairans, line 70: same Q in forward.js

src/components/button/button.js
```diff
@@ -0,0 +1,16 @@
+//@flow
+import {h} from 'preact';
+import {forwardRef} from 'preact/compat';
+
+const COMPONENT_NAME = 'Button';
```
yairans, line 5: need to export it in `src/components/index.js`

RoyBregman: Not in src/components/index.js i think. in src/components/button/index.js

src/components/forward/forward.js
RoyBregman: I don't think type button is needed for button tag

src/components/rewind/rewind.js
shalommeoded: I missed that, they were from first PR, I forgot to remove them

src/components/button/button.js
shalommeoded: Done

PR #519, opened 2020-06-23
fix(FEC-10235): Play/Pause button is not accurately centered in small player

yairans:

Since #398. 
Use `transform: translate(-50%, -50%)` and `white-space: nowrap` insteadof hardcoded `margin`

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/playback-controls/_playback-controls.scss
+++ b/src/components/playback-controls/_playback-controls.scss
@@ -4,15 +4,10 @@
   position: absolute;
   top: 50%;
   left: 50%;
-  margin-top: -32px;
-  margin-left: -51px;
-}
-
-.player.playlist {
-  .player-gui > .playback-controls,
-  .center-playback-controls {
-    margin-left: -153px;
-  }
+  transform: translate(-50%, -50%);
+  -webkit-transform: translate(-50%, -50%);
+  -ms-transform: translate(-50%, -50%);
+  white-space: nowrap;
 }
 
 .bottom-bar .playback-controls {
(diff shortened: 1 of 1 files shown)


src/components/playback-controls/_playback-controls.scss
```diff
@@ -5,13 +5,13 @@
...
   margin-top: -32px;
-  margin-left: -48px;
+  margin-left: -51px;
 }
 
 .player.playlist {
   .player-gui > .playback-controls,
   .center-playback-controls {
-    margin-left: -144px;
+    margin-left: -153px;
```
dan-ziv, line 14: `transform: translate(-50%,-50%)` instead of the hard coded margin?

yairans: it doesn't work well in playlist controls. see 
image
Since the size calculation performed before the `transform` applied. 
with `margin` it's part of the size calculation

yairans: actually works fine with `display:flex` so changed. @dan-ziv

OrenMe: see support matrix: https://caniuse.com/#feat=flexbox
Need to check IE11 specifically.

yairans: `flex` supported in ie but still looking bad in ie11 so reverted to hardcoded margin

yairans: actually works fine with `white-space: nowrap` @dan-ziv @OrenMe

PR #523, opened 2020-06-30
fix(FEC-10052): the bottom toolbar not hiding after return from background on mobile devices

dan-ziv:

Always update the hover state after player transition from paused to playing, whether it's caused by returning from background on mobile or by an API call.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/shell/shell.js
+++ b/src/components/shell/shell.js
@@ -354,6 +354,7 @@ class Shell extends Component {
     // or after an ad break
     // or in ad break
     if (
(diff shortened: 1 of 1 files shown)


src/components/shell/shell.js
```diff
@@ -354,6 +354,7 @@ class Shell extends Component {
     // or after an ad break
     // or in ad break
     if (
+      (this.props.currentState === 'playing' && prevProps.currentState === 'paused') ||
```
yairans, line 357: update the comment above

dan-ziv: done

PR #526, opened 2020-07-07
feat(FEC-9951): pip play here

RoyBregman:

As play here button in the PIP overlay was hidden under the play / pause overlay buttons when in small screens / mobile - We decided (together with Ron L and Hillel C) to remove the "Play here" button and relocate the title in the first quarter of the player instead of the center.

In addition, to make the UX more clear - the PIP button in the bottom bar has now a new state to expand back.

solves FEC-9951

### CheckLists

- [ ] Docs have been updated


src/components/picture-in-picture/picture-in-picture.js
```diff
@@ -87,13 +92,14 @@ class PictureInPicture extends Component {
     if (this.props.isPictureInPictureSupported && this.props.playerSize !== PLAYER_SIZE.EXTRA_SMALL) {
       return (
         <div className={[style.controlButtonContainer, style.pictureInPicture].join(' ')}>
-          <Tooltip label={this.props.pipText}>
+          <Tooltip label={this.props.isInPictureInPicture ? this.props.pipOpenTooltipText : this.props.pipCloseText}>
```
OrenMe, line 95: small one, but why one has tooltip in it and other doesn't?

translations/en.i18n.json
```diff
@@ -17,6 +17,8 @@
       "prev": "Prev",
       "startOver": "Start over",
       "pictureInPicture": "Picture in picture",
+      "pictureInPictureOpenToolTip": "Expand",
+      "pictureInPictureOpenAriaLabel": "Close picture in picture",
```
OrenMe, line 21: is this correct? label key is open but value is meaning close - is this correct?

src/components/picture-in-picture/picture-in-picture.js
RoyBregman: I didn't want to change existing keys. The existing one is used for both tooltip and aria label

translations/en.i18n.json
RoyBregman: I meant when PIP is open. I can add "is"

OrenMe: so I think it is more correct to call it according to the text it represents and not the state in which it is active.
Let's run this by Ron and possibly a technical writer to come with better naming and strings as I think now we

RoyBregman: done

PR #531, opened 2020-07-21
feat(FEC-10290): upgrade NPM packages

dan-ziv:

WIP

### CheckLists

- [ ] Docs have been updated


webpack.config.js
```diff
@@ -34,6 +30,9 @@ module.exports = {
   },
   devtool: 'source-map',
   plugins: plugins,
+  optimization: {
+    minimize: PROD
```
OrenMe, line 34: is this needed? just use the `--mode production` or `--mode development` in package.json scripts

src/components/video-player/video-player.js
```diff
@@ -113,7 +110,7 @@ class VideoPlayer extends Component {
   render(): React$Element<any> {
     const {videoStyles, targetId} = this.props;
 
-    return <div id={`${targetId}-video`} className={style.videoPlayer} style={videoStyles} ref={this._setRef} />;
+    return <div id={`${targetId}-video`} className={style.videoPlayer} style={videoStyles} ref={ref => (ref ? this._setRef(ref) : undefined)} />;
```
OrenMe, line 113: better not to use anonymous function for ref, can this be handled inside the setRef
also, why do we have logic in setRef? seems problematic. @yairans FYI
Can we move this to one of the lifecycle methods? if so lets handle it in another ticket.

src/components/tooltip/tooltip.js
```diff
@@ -165,7 +164,7 @@ class Tooltip extends Component {
         <span
           style={{maxWidth: props.maxWidth}}
           ref={el => {
```
OrenMe, line 166: why use this patten?
Also, see other comment about ref - better use the `createRef` preact utility for this.

src/components/share-overlay/share-overlay.js
```diff
@@ -99,7 +99,7 @@ const ShareUrl = (props: Object): React$Element<any> => {
   return (
     <div className={props.copy ? style.copyUrlRow : ''}>
       <div className={[style.formGroup, style.hasIcon, style.inputCopyUrl].join(' ')} style="width: 350px;">
-        <input tabIndex="-1" type="text" ref={c => (_ref = c)} className={style.formControl} value={props.shareUrl} readOnly />
+        <input tabIndex="-1" type="text" ref={c => (c ? (_ref = c) : undefined)} className={style.formControl} value={props.shareUrl} readOnly />
```
OrenMe, line 102: same here for usage with createRef

flow-typed/types/ui-options.js
```diff
@@ -8,5 +8,8 @@ declare type UIOptionsObject = {
   components?: ComponentsConfig,
   uiComponents?: Array<PKUIComponent>,
   translations?: {[langKey: string]: Object},
-  locale?: string
+  locale?: string,
+  disable?: boolean,
+  css?: string,
```
OrenMe, line 13: these two are actually only part of kaltura player ui-wrapper.
If we would to export types then we would have inherited from them in wrapper, and add these two there

Yuvalke:
Why do we have an event manager and fake event implementation here and in `playkit-js`?

Yuvalke:
Why don't we create a

dan-ziv:
> Why do we have an event manager and fake event implementation here and in `playkit-js`?

OrenMe:
@dan-ziv did you mean to close this?

Yuvalke:
> > Why do we have an event manager and fake event implementation here and in `playkit-js`?
> 
> @Yuvalke because we didn't want that playkit-js will be a dependency of the ui.
> I think we should create playkit-js-utils and import it from there.

@dan-ziv agree

PR #538, opened 2020-08-18
feat(FEC-10076): add support for dynamic injection

yairans:

* `UIManager` Exposes `addComponent` api which actually deligates to `PlayerAreaProvider._addNewComponentAndUpdateListeners` method (using _Render-Props_ technique)
* Remove empty array condition for removal flow 
* Extract `Fragment` children since add and remove the component behaves worried when children wrapped by a `Fragment`

### CheckLists

- [x] new unit / functional tests have been added (whenever applicable)
- [x] Docs have been updated

### DIFF
--- a/src/ui-manager.js
+++ b/src/ui-manager.js
@@ -1,4 +1,5 @@
 //@flow
+/* eslint-disable no-unused-vars */
 import {h, render} from 'preact';
 import {Provider} from 'react-redux';
 import {IntlProvider} from 'preact-i18n';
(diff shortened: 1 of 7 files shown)


src/ui-manager.js
```diff
@@ -41,6 +41,14 @@ class UIManager {
...
   _locale: string = 'en';
   _uiComponents: Array<KPUIComponent>;
+  /**
+   * Add a component dynamically
+   *
+   * @param {KPUIComponent} component - The component to add
+   * @returns {Function} - Removal function
+   * @memberof UIManager
+   */
+  addComponent: (component: KPUIComponent) => Function;
```
esakal, line 51: I would consider two things - not related to each other
1. provide a noop method as a fallback `() => { return () => {} }`. relevant also for line 181
2. expose a public method for the PlayerAreaProvider instead of passing the method to the upper component and use ref to invoke that method

both not mandatory but might be more solid

esakal (dismissed):
approving as your code seems fine. just check my comment above for possible improvements

src/ui-manager.js
yairans: 1. done
2. I tried this option. it's doesn't work since the ref points the `PlayerAreaProvider` logger wrapper

esakal (approved):

OrenMe (approved):

PR #542, opened 2020-09-07
fix(FEC-10289): controls disappear using keyboard

Yuvalke:

Issue: there are multiple timeouts that could be with different timing to handle the disappearing of UI, so there are cases that it makes the wrong behavior.
The new keyboard handler doesn't change the hover state.
Solution: handle UI disappearing in shell only
Change the hover state on new keyboard handler as it was on the previous one.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/keyboard/keyboard.js
+++ b/src/components/keyboard/keyboard.js
@@ -114,7 +114,7 @@ class Keyboard extends Component {
         this.props.player.play();
         this.props.updateOverlayActionIcon(IconType.Play);
       }
-      this.toggleHoverState();
+      this.props.updatePlayerHoverState(true);
       return {preventDefault: true, payload: true};
     },
     [KeyMap.UP]: (): KeyboardEventResult => {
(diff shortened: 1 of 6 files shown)


src/components/keyboard/keyboard.js
```diff
@@ -285,25 +284,6 @@ class Keyboard extends Component {
...
-   * @returns {void}
-   * @memberof Keyboard
-   */
-  toggleHoverState(): void {
-    if (this._hoverTimeout !== null) {
-      clearTimeout(this._hoverTimeout);
-      this._hoverTimeout = null;
-    }
-    this.props.updatePlayerHoverState(true);
-    this.props.notifyHoverChange({hover: true});
```
dan-ziv, line 301: What about the hover event? does it still be dispatched?

Yuvalke: Correct, fixed.

OrenMe (approved):

PR #544, opened 2020-09-09
fix(FEC-10481): "advertisement" title displays during bumper, when using new Ad layout

dan-ziv:

Left controls are shown although the ad is of bumper type.
This caused since the TopBar component isn't rendered again after the ad started.
Adding new state and bind a prop to the ad preset that will trigger a re-rendering after it will update.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/active-preset/active-preset.js
+++ b/src/components/active-preset/active-preset.js
@@ -16,6 +16,7 @@ const mapStateToProps = state => ({
     shell: state.shell,
     engine: {
       adBreak: state.engine.adBreak,
+      adIsBumper: state.engine.adIsBumper,
       isLive: state.engine.isLive,
       hasError: state.engine.hasError,
       isIdle: state.engine.isIdle,
--- a/src/components/engine-connector/engine-connector.js
+++ b/src/components/engine-connector/engine-connector.js
@@ -230,6 +230,7 @@ class EngineConnector extends Component {
     eventManager.listen(player, player.Event.AD_LOADED, e => {
       const ad = e.payload.ad;
       this.props.updateAdIsLinear(ad.linear);
+      this.props.updateAdIsBumper(ad.bumper);
       this.props.updateAdClickUrl(ad.clickThroughUrl);
       this.props.updateAdSkipTimeOffset(ad.skipOffset);
       this.props.updateAdSkippableState(ad.skippable);
--- a/docs/ui-components.md
+++ b/docs/ui-components.md
@@ -177,7 +177,7 @@ export class MyCustomPlugin extends KalturaPlayer.core.BasePlugin {
 }
 ```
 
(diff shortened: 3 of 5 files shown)


src/components/active-preset/active-preset.js
```diff
@@ -16,6 +16,7 @@ const mapStateToProps = state => ({
     shell: state.shell,
     engine: {
       adBreak: state.engine.adBreak,
+      adIsBumper: state.engine.adIsBumper,
```
yairans, line 19: bumper is not ad. change to `isBumper `

yairans (changes requested):

src/components/engine-connector/engine-connector.js
```diff
@@ -230,6 +230,7 @@ class EngineConnector extends Component {
     eventManager.listen(player, player.Event.AD_LOADED, e => {
```
OrenMe, line 230: Isn't the issue exactly the event listener?
The ad loaded event - we don't populate the player with ad data so getAd() will return null.

dan-ziv: That's right

OrenMe: so why not sync it to ad started?

dan-ziv: This will not fix the issue, this is just ad data props. The ad itself already initialized on ad started.

src/components/active-preset/active-preset.js
dan-ziv: You don't say! :)
Yes, but it uses our ad framework (events, presets, classes and more) so I prefer that to keep this terminology.

src/components/engine-connector/engine-connector.js
```diff
@@ -230,6 +230,7 @@ class EngineConnector extends Component {
...
       const ad = e.payload.ad;
       this.props.updateAdIsLinear(ad.linear);
+      this.props.updateAdIsBumper(ad.bumper);
```
Yuvalke, line 233: ad.bumper exists on ad_loaded from ima?

Yuvalke (changes requested):

src/components/engine-connector/engine-connector.js
OrenMe: only if we had getPendingAd API :-)

src/components/active-preset/active-preset.js
OrenMe: I actually agree with @yairans as we have two types of bumpers today - kaltura bumpers and ad bumpers which are treated as ads.
I think we should reserve the name of adBumper for future use of real ad bumpers an call this `isBumper`.

OrenMe (approved):

PR #546, opened 2020-09-15
feat(FEC-10015): support smart scrubber preview and timeline marker

yairans:
#### News:
* New `Managers` class which exposes the `registerManager(name, manager)`, `getManager(name)` and `hasManager(name)`.

#### Changes
* Add `PlayerArea` in the `Seekbar` component allows to push the cue points and to replace the default seekbar preview
* Move the seekbar preview element to `SeekBarPreview` component to be replaceable 
* Don't show the default seekbar preview when cue-point hovering (by new state `cuePointActive`)
* Ignore the seekbar mouse events when the seekbar/cue-point preview hovering (by new state `previewHoverActive`)
* Don't show the time bubble if the seekbar/cue-point preview hide it (by new state `hideTimeBubble`)
* Move the `virtualTime` to the store to be accessible from the `CuePoint`
* Add the seekbar client rect to the store to be accessible from the `CuePoint` for position calculation
* Expose some css variables to reuse in java-script 
* `PlayerArea` passes the replaced component props to the override component (used in seekbar preview overriding)
* `PlayerArea` gets new `shouldUpdate` prop to enforce updating the children (used for cue points which should get the seekbar changes)

Related to https://github.com/kaltura/kaltura-player-js/pull/359, https://github.com/kaltura/playkit-js-timeline/pull/1

Solves FEC-10015

### CheckLists

- [ ] Docs have been updated

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/playkit-js-ui/546)


esakal (changes requested):
> Reviewed 21 of 24 files at r1.
> *[Reviewable](https://reviewable.io/reviews/kaltura/playkit-js-ui/546)* status: 21 of 24 files reviewed, 3 unresolved discussions (waiting on @esakal and @yairans)

___
*[src/components/managers/managers.js, line 3 at r1](https://reviewable.io/reviews/kaltura/playkit-js-ui/546#-MHKfwyu8gH5mzJ9PMsS:-MHKfwyv9Ck_7EwUpryf:b-5nqa41) ([raw file](https://github.com/kaltura/playkit-js-ui/blob/b10afa9dafa40eaf3779b59bcccce5142b0b2f3b/src/components/managers/managers.js#L3)):*
> ```JavaScript
> //@flow
> import {TimelineManager} from './timeline-manager';
> import {UIManager} from '../../index';
> ```

this would probably lead to circle reference 
___
*[src/components/managers/timeline-manager.js, line 21 at r1](https://reviewable.io/reviews/kaltura/playkit-js-ui/546#-MHKg55Q1xxMzfaKhZDS:-MHKg55Q1xxMzfaKhZDT:b-v3j75b) ([raw file](https://github.com/kaltura/playkit-js-ui/blob/b10afa9dafa40eaf3779b59bcccce5142b0b2f3b/src/components/managers/timeline-manager.js#L21)):*
> ```JavaScript
>   addCuePoint(newCuePoint): {id: string} {
>     if (this._store.getState().engine.isLive || typeof newCuePoint.time !== 'number') {
>       return null;
> ```

this response doesn't follow the declaration because you declare that you always return an object.
either change to `{id: string} | null` or better `{ status: boolean, id: string}`

also user should know the reason it fails. if in live it will do nothing, if the cuepoint is invalid it will always fail. so consider `{ errorCode?: number, data?: { id: string} }`
___
*[src/components/managers/timeline-manager.js, line 76 at r1](https://reviewable.io/reviews/kaltura/playkit-js-ui/546#-MHKh5M89K8Bd-pbJSPj:-MHKh5M89K8Bd-pbJSPk:bhe2v6e) ([raw file](https://github.com/kaltura/playkit-js-ui/blob/b10afa9dafa40eaf3779b59bcccce5142b0b2f3b/src/components/managers/timeline-manager.js#L76)):*
> ```JavaScript
>       replaceComponent: 'SeekBarPreview'
>     });
>     this._store.dispatch(seekbarActions.updateHideTimeBubble(false));
> ```

I'm not sure what this means. But it seems that you can have multiple preview instances while you change a shared state based on this method, is this reasonable?

yairans (commented):
> *[Reviewable](https://reviewable.io/reviews/kaltura/playkit-js-ui/546)* status: 21 of 24 files reviewed, 3 unresolved discussions (waiting on @esakal and @yairans)

___
*[src/components/managers/managers.js, line 3 at r1](https://reviewable.io/reviews/kaltura/playkit-js-ui/546#-MHKfwyu8gH5mzJ9PMsS:-MHLBIzUD2Q1ycPl32gL:bt1r8h3) ([raw file](https://github.com/kaltura/playkit-js-ui/blob/b10afa9dafa40eaf3779b59bcccce5142b0b2f3b/src/components/managers/managers.js#L3)):*
<details><summary><i>Previously, esakal (Eran Sakal) wrote…</i></summary><blockquote>

this would probably lead to circle reference 
</blockquote></details>

I need it for typing, seems like webpack is OK with that.
___
*[src/components/managers/timeline-manager.js, line 76 at r1](https://reviewable.io/reviews/kaltura/playkit-js-ui/546#-MHKh5M89K8Bd-pbJSPj:-MHLBpyIEYHHxJg-enkj:b-x71utf) ([raw file](https://github.com/kaltura/playkit-js-ui/blob/b10afa9dafa40eaf3779b59bcccce5142b0b2f3b/src/components/managers/timeline-manager.js#L76)):*
<details><summary><i>Previously, esakal (Eran Sakal) wrote…</i></summary><blockquote>

I'm not sure what this means. But it seems that you can have multiple preview instances while you change a shared state based on this method, is this reasonable?
</blockquote></details>

This action is for when the seekbar preview (not cue point preview) wants to hide the time bubble. 
There is only one seekbar preview so it's reasonable.

src/components/managers/timeline-manager.js
```diff
@@ -0,0 +1,123 @@
...
+    this._counter = 0;
+  }
+
+  /**
+   * @param {CuePointOptionsObject} newCuePoint - The cue point options
+   * @return {null|{id: string}} - An object contains the cue point id
+   */
+  addCuePoint(newCuePoint: CuePointOptionsObject): {id: string} | null {
+    if (this._store.getState().engine.isLive) {
+      logger.warn('Impossible to add cue points while LIVE playback');
```
yairans, line 37: fix logger. see https://github.com/kaltura/playkit-js-ui/pull/552

src/components/managers/managers.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+class Managers {
+  _timeline: TimelineManager;
+
+  /**
+   * @constructor
+   * @param {UIManager} uiManager - The UI manager.
+   * @param {any} store - The store.
+   */
+  constructor(uiManager: UIManager, store: any) {
```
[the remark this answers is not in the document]
yairans: done

src/components/managers/managers.js
```diff
@@ -0,0 +1,64 @@
...
+      Managers._logger.debug(`${name} manager registered`);
+    }
+  }
+
+  /**
+   *
+   * @param {string} name - the manager name
+   * @returns {Object} - the manager object
+   */
+  getManager(name: string): Object | void {
```
dan-ziv, line 42: Create enums for the names and type it.

src/components/cue-point/cue-point.scss
```diff
@@ -0,0 +1,34 @@
...
+  margin: -12px 0;
+
+  &:hover {
+    .cue-point-preview-container {
+      display: block;
+    }
+  }
+}
+
+.cue-point {
```
dan-ziv, line 16: define this class inside `cue-point-container`

src/components/managers/managers.js
yairans: the `managers` is open to expanding by register manager from outside. this couldn't be in enum.

src/components/cue-point/cue-point.scss
yairans: I don't want to be specific. this makes the default class be stronger than a custom class provided by the app (means the app will need to add `!important`).

PR #549, opened 2020-10-05
fix(FEC-10481): "advertisement" title displays during bumper, when using new ad layout 

dan-ziv:

`TopBar` component isn't re-rendered when its `leftControls` prop change to `undefined`.
Solution: Manage the `AdNotice` component inside the component itself by binding the `adIsBumper` state to it.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/active-preset/active-preset.js
+++ b/src/components/active-preset/active-preset.js
@@ -16,7 +16,6 @@ const mapStateToProps = state => ({
     shell: state.shell,
     engine: {
       adBreak: state.engine.adBreak,
-      adIsBumper: state.engine.adIsBumper,
       isLive: state.engine.isLive,
       hasError: state.engine.hasError,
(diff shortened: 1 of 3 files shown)


OrenMe:
If possible we would like to avoid connecting nested presentational components to redux store.
Is there another way? Why is preset not being re-rendered on change of props?

dan-ziv:
@OrenMe the preset is re-rendered, the `TopBar` component isn't.
Other solution I can think of at the preset level is something that will look like that:
```
isBumper(props) ? <TopBar disabled={true} leftControls={undefined} /> :  <TopBar disabled={true} leftControls={<AdNotice />} />
```

OrenMe:
But this leaks the top bar children logic to top bar.
Please check how we cause render params to cause rerender of parent.
@esakal maybe you can advise?

PR #553, opened 2020-10-21
fix(FEC-10503): watermark doesn't show up correctly on top and doesn't clickable.

Yuvalke:

Issue: watermark wasn't clickable and respect empty spaces on bars(top and bottom), the top bar didn't make the watermark to move.
Solution: watermark clickable and use the all empty space valid from top and bottom.
Bars and interactive area are located directly under GUI area so new children of GUI area will be under the bars and interactive area in the hierarchy and every use of  GUI area should be with MAX 2 child one is the content of GUI area and second is bars and interactive area when it exists.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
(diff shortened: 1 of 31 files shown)


yairans (changes requested):

yairans (approved):
LGTM

PR #560, opened 2020-11-09
feat(FEC-10527): add RW/FF controls to live preset

RoyBregman:

Added the new controls to the live preset (only for DVR)
and added some checks in the FF and RW buttons when on the edges to avoid jumps in timeline

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/rewind/rewind.js
+++ b/src/components/rewind/rewind.js
@@ -26,10 +26,10 @@ export const REWIND_DEFAULT_STEP = 10;
  * @returns {Object} - mapped state to this component
  */
 const mapStateToProps = state => ({
-  isDvr: state.engine.isDvr
+  isDvr: state.engine.isDvr,
+  isLive: state.engine.isLive
 });
 
-@connect(mapStateToProps)
 /**
  * Rewind component
  *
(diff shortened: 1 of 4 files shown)


src/ui-presets/live.js
```diff
@@ -86,6 +99,8 @@ class LiveUI extends Component {
                     leftControls={
                       <Fragment>
                         <PlaybackControls />
+                        {this.props.isDvr ? <Rewind step={10} /> : null}
+                        {this.props.isDvr ? <Forward step={10} /> : null}
```
OrenMe, line 103: remove state from here, return undefined from components if live and not dvr

src/components/rewind/rewind.js
```diff
@@ -44,7 +55,10 @@ class Rewind extends Component {
     const step = this.props.step || REWIND_DEFAULT_STEP;
     const from = this.props.player.currentTime;
     if (this.props.player.currentTime - step < 0) {
-      to = 0;
+      // In dvr when close to beginning dont rewind
+      if (!this.props.isDvr) {
```
OrenMe, line 59: doesn't this mean that rewind won't take us all the way back? it will always remain a `REWIND_DEFAULT_STEP` from actual 0?

OrenMe (changes requested):

src/components/rewind/rewind.js
RoyBregman: It will some margin < REWIND_DEFAULT_STEP but this is better behaviour than move to the actual 0 which causes jumps of a whole segment sometimes

OrenMe (approved):

PR #569, opened 2020-12-24
fix(FEC-10670): Add missing class names

dan-ziv:

Remove unused class names so no redundant spaces will be generated in our HTML.

Note: other solution suggested such as adding an empty class container doesn't actually adding it to the HTML so the issue stays the same.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/language/language.js
+++ b/src/components/language/language.js
@@ -204,7 +204,7 @@ class Language extends Component {
     return (
       <div
         ref={c => (c ? (this._controlLanguageElement = c) : undefined)}
-        className={[style.controlButtonContainer, style.controlLanguage].join(' ')}>
+        className={style.controlButtonContainer}>
         <Tooltip label={this.props.buttonLabel}>
           <Button
             tabIndex="0"
--- a/src/components/language/language.js
+++ b/src/components/language/language.js
@@ -18,6 +18,7 @@ import {KeyMap} from 'utils/key-map';
 import {withKeyboardEvent} from 'components/keyboard';
 import {Tooltip} from 'components/tooltip';
 import {Button} from 'components/button';
+import {controlButton} from 'utils/control-button';
 
 /**
  * mapping state to props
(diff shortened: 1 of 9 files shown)


OrenMe:
Need to find a solution which will enable a dedicated class name for each top level button in our control bar

OrenMe:
@dan-ziv please re-review

dan-ziv:
@OrenMe done

src/components/language/language.js
```diff
@@ -202,9 +203,7 @@ class Language extends Component {
   renderAll(audioOptions: Array<Object>, textOptions: Array<Object>): React$Element<any> {
     const portalSelector = `#${this.props.player.config.targetId} .overlay-portal`;
     return (
-      <div
-        ref={c => (c ? (this._controlLanguageElement = c) : undefined)}
```
OrenMe, line 206: I think this is used for A11y, @RoyBregman, right?

src/utils/control-button.js
```diff
@@ -0,0 +1,31 @@
...
+     * render component
+     *
+     * @returns {React$Element} - component element
+     * @memberof ControlButtonComponent
+     */
+    render(): React$Element<any> {
+      return (
+        <div
+          ref={c => (c ? (this._buttonElement = c) : undefined)}
+          className={[style.controlButtonContainer, `${__CSS_MODULE_PREFIX__}-${componentName.toLowerCase()}-control`].join(' ')}>
```
OrenMe, line 23: I think this might be a bit bigger than what I initially thought.
Need to take into account - sometimes the styles are indeed coming from the styles.scss file so we don't necessirily want to give up on this.
Secondly, this is adding another layer of wrapping around our components, which can easily be done with a small helper function instead.

Original issue is due to fact that adding an empty class gets minimized by the css loader minifier.
https://stackoverflow.com/questions/42823494/retaining-empty-unused-classes-with-css-loader
https://github.com/css-modules/css-modules/issues/140

dan-ziv: There's no problem with styles comes from `style.scss` (for example like in `Volume` component). The only thing the decorator does is to append the appropriate class name on the top of the control button element. If styles are persist it will apply the settings from the file.

OrenMe: but we are adding an additional div here, not the expected behviour.
I would even go to simplyfing it to add hardcoded string values where ever a scss class doesn't exist.

PR #572, opened 2021-01-11
fix(FEC-10670): Add missing class names

dan-ziv:

Create `ControlButton` component and composite it all control buttons components. 

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/volume/_volume.scss
+++ b/src/components/volume/_volume.scss
@@ -1,6 +1,6 @@
 @import '~styles/variables';
 
-.control-button-container.volume-control {
+.control-button-container.control-volume {
   display: inline-block !important;
   &.hover {
     .volume-control-bar {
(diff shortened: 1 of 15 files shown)


src/components/volume/_volume.scss
```diff
@@ -1,6 +1,6 @@
 @import '~styles/variables';
 
-.control-button-container.volume-control {
+.control-button-container.control-volume {
```
yairans, line 3: This is breaking change. There are customers override the volume style by this class

yairans (dismissed):

src/components/volume/_volume.scss
dan-ziv: Fixed.
BTW, checked on customers sites, they are all overriding the icon classes (`playkit-icon-volume` for example) so this ain't break anything. But anyway, for us to be on the safe side, fixed.

yairans (approved):

PR #579, opened 2021-02-14
feat(FEC-10667): control-bar auto-hide timeout configuration

Yuvalke:

Issue: unable to change our default for hiding the control bar.
Solution: add a new config called controlBarHoverTimeout which will pass the duration of timeout the user wants for this action.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -16,7 +16,7 @@ var uiManager = new playkit.ui.UIManager(player, config);
   targetId: string,
   debugActions?: boolean, // optional
   forceTouchUI?: boolean, // optional
-  controlBarHoverTimeout?: number, // optional
+  hoverTimeout?: number, // optional
   logger?: loggerType, // optional
   components?: Object, // optional
   uiComponents: Array<Object>, //optional
@@ -61,7 +61,7 @@ var uiManager = new playkit.ui.UIManager(player, config);
 
 ##
 
-> ### config.controlBarHoverTimeout
+> ### config.hoverTimeout
 >
 > ##### Type: `number`
 >
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -16,6 +16,7 @@ var uiManager = new playkit.ui.UIManager(player, config);
   targetId: string,
   debugActions?: boolean, // optional
   forceTouchUI?: boolean, // optional
+  controlBarHoverTimeout?: number, // optional
   logger?: loggerType, // optional
   components?: Object, // optional
   uiComponents: Array<Object>, //optional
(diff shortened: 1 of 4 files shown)


src/components/shell/shell.js
```diff
@@ -325,11 +326,13 @@ class Shell extends Component {
    */
   _startHoverTimeout(): void {
     this._clearHoverTimeout();
-    this.hoverTimeout = setTimeout(() => {
-      if (this._canEndHoverState()) {
-        this._updatePlayerHover(false);
-      }
-    }, this.props.hoverTimeout || CONTROL_BAR_HOVER_DEFAULT_TIMEOUT);
```
yairans, line 332: so need to delete `CONTROL_BAR_HOVER_DEFAULT_TIMEOUT`

yairans: better to check why it exported at all

src/reducers/config.js
```diff
@@ -9,6 +10,7 @@ export const types = {
 
 export const initialState = {
   forceTouchUI: false,
+  controlBarHoverTimeout: CONTROL_BAR_HOVER_DEFAULT_TIMEOUT,
```
yairans, line 13:
```suggestion
  controlBarHoverTimeout: 3000,
```

yairans (changes requested):

src/components/shell/shell.js
Yuvalke: Maybe a customer who adding custom UI can take the shell and import this data from the shell.
It'll break who use this constant

docs/configuration.md
```diff
@@ -16,6 +16,7 @@ var uiManager = new playkit.ui.UIManager(player, config);
   targetId: string,
   debugActions?: boolean, // optional
   forceTouchUI?: boolean, // optional
+  controlBarHoverTimeout?: number, // optional
```
OrenMe, line 19: the config is for entire UI so let's find a more suitable name then `controlBarHoverTimeout`
Maybe `userHoverTimeout` or `userActionTimeout`

src/components/shell/shell.js
Yuvalke: It was added here - https://kaltura.atlassian.net/browse/FEC-7373
@yairans  I don't think we're able to know if it was on purpose or not...(it looks like it's not)

docs/configuration.md
Yuvalke: `hoverTimeout` ?

OrenMe: that was the original one that was never used internally, and I have no objection for it.

docs/configuration.md
```diff
@@ -60,6 +61,18 @@ var uiManager = new playkit.ui.UIManager(player, config);
...
 
+> ### config.hoverTimeout
+>
+> ##### Type: `number`
+>
+> ##### Default: `3000`
+>
+> ##### Description: Defines the timeout for control bar hover, 0 - don't hide.
+>
+> Useful for applications that wants to set different value for bar hover display.
```
OrenMe, line 72:
```suggestion
> ##### Description: Defines the timeout for control bar hover, 0 - always show.
>
> Useful for applications that wants to set different hover timeout duration for player controls.
```

src/components/shell/shell.js
yairans: you can remove it as no one can import it outside the player
see https://github.com/kaltura/playkit-js-ui/blob/master/src/components/shell/index.js
we exporting `Shell` only (for all we have - `export *`)

yairans (changes requested):

yairans (approved):

OrenMe (approved):

PR #583, opened 2021-03-08
feat(FEC-10941): Use In-Stream DASH thumbnails on the timeline

dan-ziv:

Move UI to work with thumbnail API instead of thumbnail config. 

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/seekbar-preview/seekbar-preview.js
+++ b/src/components/seekbar-preview/seekbar-preview.js
@@ -2,6 +2,9 @@
 import style from '../../styles/style.scss';
 import {h, Component} from 'preact';
 import {connect} from 'react-redux';
+import {withPlayer} from 'components';
+
+const FRAME_PREVIEW_IMG_CONTAINER_OFFSET: number = 4;
 
 /**
  * mapping state to props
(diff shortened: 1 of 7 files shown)


src/components/seekbar-preview/seekbar-preview.js
```diff
@@ -18,41 +21,47 @@ const mapStateToProps = state => ({
...
+   * @param {Object} thumbnailInfo  - The thumbnail info data.
+   * @returns {string} - The css style string.
+   * @private
    */
-  getThumbSpriteOffset(): string {
-    const percent = this.props.virtualTime / this.props.duration;
-    const sliceIndex = Math.ceil(this.props.thumbsSlices * percent);
-    return -(sliceIndex * this.props.thumbsWidth) + 'px 0px';
+  _getFramePreviewImgStyle(thumbnailInfo: Object): string {
+    let framePreviewImgStyle = `background: url(${thumbnailInfo.url});`;
```
yairans, line 34: it was `background-image: url....` what's the difference?

yairans (changes requested):

src/components/seekbar-preview/seekbar-preview.js
dan-ziv: Just different syntax since we don't need previous styles anymore.
In a background property you can add `background-color`, `repeat`, `no-repeat` and other image attributes, but in the background-image property you are only allowed to add image.

```
background-image: url("");
background-repeat: no-repeat;
background-position: right top;
```

and in background property you can do in one line all these

`background: #ccc url("") no-repeat;`

yairans (approved):

PR #589, opened 2021-04-05
feat(FEC-10817): expose the current time of the video element in live streams

dan-ziv:

expose the current time of the video element in live streams

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/engine-connector/engine-connector.js
+++ b/src/components/engine-connector/engine-connector.js
@@ -78,7 +78,7 @@ class EngineConnector extends Component {
     });
 
     eventManager.listen(player, player.Event.TIME_UPDATE, () => {
-      this.props.updateCurrentTime(player.currentTime);
+      this.props.updateCurrentTime(player.isLive() ? player.liveTime : player.currentTime);
     });
 
     eventManager.listen(player, player.Event.DURATION_CHANGE, () => {
--- a/src/components/seekbar-live-playback-container/seekbar-live-playback-container.js
+++ b/src/components/seekbar-live-playback-container/seekbar-live-playback-container.js
@@ -49,7 +49,7 @@ class SeekBarLivePlaybackContainer extends Component {
   componentDidMount() {
     this.props.eventManager.listen(this.props.player, this.props.player.Event.TIME_UPDATE, () => {
       if (!this.props.isDraggingActive) {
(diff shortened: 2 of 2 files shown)


src/components/live-tag/live-tag.js
```diff
@@ -77,13 +77,14 @@ class LiveTag extends Component {
...
    * render live tag component
    *
-   * @param {*} props - component props
    * @returns {React$Element} component element
    * @memberof LiveTag
    */
-  render(props: any): React$Element<any> {
+  render(): React$Element<any> {
     const tagStyleClass = [style.liveTag];
-    if (props.isDvr && !this.isOnLiveEdge()) tagStyleClass.push(style.nonLivePlayhead);
```
yairans, line 86: why `isDvr` removed? if needed so remove it from mapStateToProps as well

src/components/seekbar/seekbar.js
```diff
@@ -542,7 +542,7 @@ class SeekBar extends Component {
    */
   render(props: any, state: Object): React$Element<any> {
     const virtualProgressWidth = `${(props.virtualTime / props.duration) * 100}%`;
-    const progressWidth = `${(props.currentTime / props.duration) * 100}%`;
+    const progressWidth = `${props.forceFullProgress ? 100 : (props.currentTime / props.duration) * 100}%`;
```
yairans, line 545: why is needed?

yairans (changes requested):

src/components/seekbar/seekbar.js
dan-ziv: Now the live time is linear and duration is always bigger than the current time. For that reason you will see unfulfilled progress on the progress bar's edge. When we are on live edge this has no meaning so we would like to see full progress.

src/components/live-tag/live-tag.js
dan-ziv: If you are out-of-sync (paused for example) also on live without DVR, we would like to see indication on the live tag.

yairans: this is new behavior. how is related to this change?

dan-ziv: Now the calculation is different so we can support it. Like before it's a bug.

yairans (approved):

yairans (approved):

PR #590, opened 2021-04-06
fix(FEC-11128): X button on floating player not working

lianbenjamin:

when an ad is playing, the dismiss button is showing but not working.
made all left-controls and right-controls children clickthrough and excepted ad-notice.

Solves FEC-11128

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/top-bar/_top-bar.scss
+++ b/src/components/top-bar/_top-bar.scss
@@ -1,5 +1,3 @@
-@import '../ad-notice/ad-notice';
-
 .player .top-bar {
   background: linear-gradient(0deg, rgba(0, 0, 0, 0) 0%, rgba(0, 0, 0, 0.6) 100%);
   color: #fff;
@@ -30,10 +28,7 @@
     &:empty {
       margin: 0;
     }
-
-    .ad-notice {
-      pointer-events: none;
-    }
+    pointer-events: none;
   }
   .right-controls {
     float: right;
@@ -42,6 +37,7 @@
     &:empty {
       margin: 0;
     }
+    pointer-events: none;
 
     .control-button-container {
       margin: 0 6px;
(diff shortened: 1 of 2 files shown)


RoyBregman (approved):

src/components/top-bar/_top-bar.scss
```diff
@@ -28,6 +30,10 @@
     &:empty {
       margin: 0;
     }
+
+    .ad-notice {
```
yairans, line 34: ad notice has its own css file. do it there

src/components/top-bar/_top-bar.scss
```diff
@@ -45,6 +51,11 @@
       }
     }
   }
+
+  .left-controls *,
+  .right-controls * {
+    pointer-events: auto;
```
yairans, line 57: the top-bar itself it also a container so add this in it

yairans (changes requested):

src/components/top-bar/_top-bar.scss
yairans: with scss you can do like:
```
  &.disabled {
    pointer-events: none;
    * {
       pointer-events: auto;
    }
  }```

src/components/top-bar/_top-bar.scss
lianbenjamin: @yairans it's not working when adding this to ad-notice.scss. the adding of pointer-events:auto to top-bar's children is stronger. we needed to add another level to make this one stronger.

src/components/top-bar/_top-bar.scss
RoyBregman: We tried first what you suggested but it doesn't fix the issue correctly. The problem is that the left and right controller also have dimensions due to its children, so it is also grabbing the clicks. That is why we lowered it another level

yairans (approved):

PR #599, opened 2021-05-09
fix(FEC-10811): floating show two bottom bar for second

Yuvalke:

Issue: the first event of resize/scroll isn't taken into account only the last one in the sequence.
Solution: trigger the function on the leading edge instead of the trail.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/utils/debounce.js
+++ b/src/utils/debounce.js
@@ -19,11 +19,15 @@ export const debounce: Function = (origFunc: Function, time: number, immediate: 
 
   return (...args) => {
     const context = this;
-    if (immediate && !timeout) origFunc.apply(context, args);
+    if (immediate && !timeout) {
+      origFunc.apply(context, args);
+    }
     clearTimeout(timeout);
     timeout = setTimeout(function () {
       timeout = null;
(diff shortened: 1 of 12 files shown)


RoyBregman (approved):

src/utils/debounce.js
```diff
@@ -13,13 +14,16 @@
  *  }, ON_WINDOW_RESIZE_DEBOUNCE_DELAY)
  * );
  */
-export const debounce: Function = (origFunc: Function, time: number) => {
+export const debounce: Function = (origFunc: Function, time: number, immediate: boolean = true) => {
   let timeout;
 
   return (...args) => {
+    const context = this;
+    if (immediate && !timeout) origFunc.apply(context, args);
```
OrenMe, line 22: Do not use inline ifs without curly brackets.
@yairans letw make sure we have lint rule to make sure this is not possible

OrenMe (changes requested):

OrenMe:
@Yuvalke sid you mean to write "isn't taken into account"?

yairans (approved):

PR #613, opened 2021-06-22
feat(FEC-10785): add support to remove ui element

RoyBregman:

Until now it was possible to add new component or to add and replace existing components.
There was no official way to remove an existing ui component.
The new way to that now is to use replaceComponent with the existing component and add the special "remove" string in the get property instead of a new component.

solves FEC-10785
changes were done KP repo also - https://github.com/kaltura/kaltura-player-js/pull/460

### CheckLists

- [X] Docs have been updated

### DIFF
--- a/src/components/player-area/player-area-provider.js
+++ b/src/components/player-area/player-area-provider.js
@@ -82,7 +82,7 @@ class PlayerAreaProvider extends Component {
   _validateComponentData = componentData => {
     // we keep option `container` for backward compatibility. documentation are showing `area` property
     const hasAreaProperty = componentData.container || componentData.area;
-    if ((!componentData.get && !componentData.removeComponent) || !componentData.presets || !hasAreaProperty) {
+    if (!componentData.get || !componentData.presets || !hasAreaProperty) {
       this.props.logger.warn(
         `component data with label '${componentData.label || ''}' is invalid (did you remember to set 'get', 'presets' and 'area'?)`
       );
--- a/src/components/player-area/player-area-provider.js
+++ b/src/components/player-area/player-area-provider.js
@@ -82,7 +82,7 @@ class PlayerAreaProvider extends Component {
   _validateComponentData = componentData => {
     // we keep option `container` for backward compatibility. documentation are showing `area` property
     const hasAreaProperty = componentData.container || componentData.area;
-    if (!componentData.get || !componentData.presets || !hasAreaProperty) {
+    if ((!componentData.get && !componentData.removeComponent) || !componentData.presets || !hasAreaProperty) {
       this.props.logger.warn(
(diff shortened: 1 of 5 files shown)


src/components/player-area/player-area-provider.js
```diff
@@ -82,7 +82,7 @@ class PlayerAreaProvider extends Component {
   _validateComponentData = componentData => {
     // we keep option `container` for backward compatibility. documentation are showing `area` property
     const hasAreaProperty = componentData.container || componentData.area;
-    if (!componentData.get || !componentData.presets || !hasAreaProperty) {
+    if ((!componentData.get && !componentData.removeComponent) || !componentData.presets || !hasAreaProperty) {
```
yairans, line 85: change the log below accordingly

yairans (changes requested):

src/components/player-area/player-area-provider.js
OrenMe, line 85: I think that because until now our semantics didn't contain **action + Component** with component suffix we should just call it remove and not removeComponent.

OrenMe (dismissed):

yairans (dismissed):

src/components/player-area/player-area-provider.js
RoyBregman: Current message also describes our case. If we need to make it more specific then we need to separate the if statement

RoyBregman: Today we have also - replaceComponent, beforeComponent, afterComponent. I kept the same semantics

OrenMe: u r totally right, not sure why I thought the other way around :-) so please disregard.

OrenMe: but now I have another thought on this.
I don't see removeComponent as same as before|After|replace exactly.
All of the above talk about positional operations, e.g., here to place a component, while remove is not.
Wouldn't it make more sense to re-use replaceComponent and add a get placeholder for a component called None or something that says **_no component_**  or keep just the replaceComponent and if you have replaceComponent and no get then it means to remove it?

Feels redundant to add more config options and make the object fat where we already have all of the current config options.

RoyBregman: To me it felt more natural to have a new prop called removeComponent.
But I guess with good documentation it is very possible to do it like you suggested

OrenMe: you know how I personally feel about new APIs or config options - we should always consider if it is necessary as new APIs or not as new ones are an obligation by us and cannot be changed/broken easily after they are committed.
In any case, this is great addition - also need to be able to define this from studio and uiconf which is very useful as we saw in VE.

yairans (approved):

PR #620, opened 2021-07-07
fix(FEC-11393): unmute button are cut off when not on hover/there's no top bar

RoyBregman:

When we hide the top bar we put margin-top -60px on it. But if the top bar is empty its height is just 16px and therefore the top bar is moved too high and the interactive area is starting too high also.
The fix is to enforce that when the top is hidden it will be exactly 60px.
I also removed the 16px margin in the children when empty and now it is handled by min-height: 16px
I also fixed some issue caused by css ranking in sm size

solves FEC-11393

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/top-bar/_top-bar.scss
+++ b/src/components/top-bar/_top-bar.scss
@@ -6,8 +6,10 @@
   visibility: hidden;
   position: relative;
   height: auto;
+  min-height: #{$top-bar-max-height}px;
   margin-top: #{-$top-bar-max-height}px;
-  transition: #{$hover-animation-time}ms visibility ease-in-out, #{$hover-animation-time}ms margin-top ease-in-out;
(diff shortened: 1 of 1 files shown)


src/components/top-bar/_top-bar.scss
```diff
@@ -90,17 +93,24 @@
...
+    }
   }
   .top-bar-area {
     width: calc(100% - #{2 * $gui-small-gutter}px);
     margin: #{$top-bar-top-bottom-gutter}px #{$gui-small-gutter}px;
+    &:empty {
+      margin: 0 0 #{$gui-gutter}px 0;
+    }
   }
 }
```
RoyBregman, line 116: This solves another bug in top bar - padding is not used anymore since we added margins to the children and also the margin css was overridden from the regular size hover margin.

yairans (approved):

PR #621, opened 2021-07-13
feat(FEC-10666): add HD/4K indication in quality selector

JonathanTGold:
### Context
Screen Shot 2021-07-13 at 12 15 01

Currently, the quality selector lists all resolutions that appear in the manifest, taking the height from the resolution and appending a 'P'.
Qualities would look like: 360P, 720P, 1080P, 1440P, etc

However, today's viewers do not know what 720p or 1440p means. They know SD, HD and 4K (and soon - 8K).
We wish to add a simple indication that would allow millennials to know the quality they are watching.

### Requirements
For resolutions, where height is between 720 (included) and 2160 (excluded) add an HD indication.
For resolutions, where height is between 2160 (included) and 4320 (excluded) add a 4K indication.
For resolutions, where height is above 4320 (included) add a 8K indication.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/badge/_badge.scss
+++ b/src/components/badge/_badge.scss
@@ -1,0 +1,13 @@
+.badge {
+  height: 10px;
+  width: 15px;
+  background-color: white;
+  color: black;
+  border-radius: 1px;
+  font-size: 9px;
+  text-transform: uppercase;
+  margin-left: 3px;
+  line-height: 10px !important;
+  text-align: center;
+  vertical-align: top !important;
+}
--- a/src/components/badge/_badge.scss
+++ b/src/components/badge/_badge.scss
@@ -8,7 +8,7 @@
   text-transform: uppercase;
   margin-left: 3px;
   margin-top: 2px;
-  align-self: start;
+  align-self: flex-start;
   display: flex;
   justify-content: center;
   align-items: center;
--- a/src/components/badge/_badge.scss
+++ b/src/components/badge/_badge.scss
@@ -1,7 +1,5 @@
-$base-height: 10px;
-
 .badge {
-  height: $base-height;
+  height: 10px;
   width: 15px;
   background-color: white;
   color: black;
@@ -9,7 +7,9 @@ $base-height: 10px;
   font-size: 9px;
   text-transform: uppercase;
   margin-left: 3px;
-  line-height: $base-height !important;
-  text-align: center;
-  vertical-align: top !important;
+  margin-top: 2px;
+  align-self: start;
+  display: flex;
+  justify-content: center;
+  align-items: center;
 }
--- a/src/components/badge/_badge.scss
+++ b/src/components/badge/_badge.scss
@@ -1,5 +1,7 @@
+$base-height: 10px;
+
 .badge {
-  height: 10px;
+  height: $base-height;
   width: 15px;
   background-color: white;
   color: black;
@@ -7,7 +9,7 @@
   font-size: 9px;
   text-transform: uppercase;
   margin-left: 3px;
-  line-height: 10px !important;
+  line-height: $base-height !important;
   text-align: center;
   vertical-align: top !important;
 }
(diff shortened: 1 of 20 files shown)


src/components/badge/_badge.scss
```diff
@@ -0,0 +1,15 @@
...
+.badge {
+  height: $base-height;
+  width: 15px;
+  background-color: white;
+  color: black;
+  border-radius: 1px;
+  font-size: 9px;
+  text-transform: uppercase;
+  margin-left: 3px;
+  line-height: $base-height !important;
```
RoyBregman, line 12: don't use !important. It is bad practise. If you need to override other css then add more level of selectors to increase css rank

src/components/menu/menu.js
```diff
@@ -264,6 +264,7 @@ class MenuItem extends Component {
         onClick={this.onClick}
         onKeyDown={this.onKeyDown}>
         <span>{props.data.label}</span>
+        <span>{props.data.badge ? props.data.badge : null}</span>
```
RoyBregman, line 267: If think the span is redundant. If it needed to be wrapped then wrap it with fragment. But I think this should also work:
```suggestion
        {props.data.badge ? props.data.badge : null}
```

src/components/settings/settings.js
```diff
@@ -243,6 +244,25 @@ class Settings extends Component {
...
+   * Prepares the badge of the quality option according to the height of its resolution
+   *
+   * @param {number} videoTrackHeight - video track
+   * @returns {Component<Badge>}
+   * @memberof Settings
+   */
+  getBadge(videoTrackHeight: number): Component<Badge> {
+    let badgeContent = '';
+    if (videoTrackHeight >= 720 && videoTrackHeight < 2160) {
+      badgeContent = 'HD';
```
RoyBregman, line 257: Should this be translated in other languages? If so we need to take out to localization file

src/components/settings/settings.js
```diff
@@ -243,6 +244,25 @@ class Settings extends Component {
...
+  /**
+   * Prepares the badge of the quality option according to the height of its resolution
+   *
+   * @param {number} videoTrackHeight - video track
+   * @returns {Component<Badge>}
+   * @memberof Settings
+   */
+  getBadge(videoTrackHeight: number): Component<Badge> {
+    let badgeContent = '';
+    if (videoTrackHeight >= 720 && videoTrackHeight < 2160) {
```
RoyBregman, line 256: Take the numbers out to some const

src/components/settings/settings.js
JonathanTGold: no, i asked ron

RoyBregman (commented):
LGTM, please recheck the feature in Safari, FF, IE, mobile after the new css fix of the dropdown

OrenMe:
please pay attention that we have different design on mobile devices and also for small screen.
Figma design didn't take this into account so check how it looks and maybe get UX designer involved.

src/components/settings/settings.js
```diff
@@ -243,6 +250,25 @@ class Settings extends Component {
...
+  getBadge(videoTrackHeight: number): Component<Badge> {
+    let badgeContent = '';
+    if (videoTrackHeight >= HeightResolution.HD && videoTrackHeight < HeightResolution.UHD_4K) {
+      badgeContent = 'HD';
+    } else if (videoTrackHeight >= HeightResolution.UHD_4K && videoTrackHeight < HeightResolution.UHD_8K) {
+      badgeContent = '4K';
+    } else if (videoTrackHeight >= HeightResolution.UHD_8K) {
+      badgeContent = '8K';
+    }
+    return videoTrackHeight >= HeightResolution.HD ? <Badge content={badgeContent} /> : null;
```
yairans, line 269:
```suggestion
    return badgeContent ? <Badge content={badgeContent} /> : null;
```

yairans (approved):

PR #633, opened 2021-08-23
fix:(FEC-10962): The player's skin should not disappear after clicking on the control bar's buttons (even when the mouse wasn't moved for a few seconds)

JonathanTGold:
issue: hover state timer of 3 seconds restarting just on mouse-move event
fix: reset the timer on general ui-click event
solves: FEC-10962

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/bottom-bar/_bottom-bar.scss
+++ b/src/components/bottom-bar/_bottom-bar.scss
@@ -80,7 +80,7 @@
 }
 
 .player.casting,
-.player.metadata-loaded:hover,
+.player.metadata-loaded.hover,
 .player.state-paused,
 .player.state-idle,
 .player.ad-break,
--- a/src/components/bottom-bar/_bottom-bar.scss
+++ b/src/components/bottom-bar/_bottom-bar.scss
@@ -80,7 +80,7 @@
 }
 
 .player.casting,
-.player.metadata-loaded.hover,
+.player.metadata-loaded:hover,
 .player.state-paused,
 .player.state-idle,
 .player.ad-break,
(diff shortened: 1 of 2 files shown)


src/components/bottom-bar/_bottom-bar.scss
```diff
@@ -80,7 +80,7 @@
 }
 
 .player.casting,
-.player.metadata-loaded.hover,
+.player.metadata-loaded:hover,
```
yairans, line 83: 1. the player has a timeout for hover toggling. see https://github.com/kaltura/playkit-js-ui/blob/master/src/components/shell/shell.js#L322
it won't work with this change
2. please note the top-bar has these css rulls too

yairans (changes requested):

src/components/shell/shell.js
```diff
@@ -226,6 +226,7 @@ class Shell extends Component {
     eventManager.listen(this._playerResizeWatcher, FakeEvent.Type.RESIZE, debounce(this._onWindowResize, ON_PLAYER_RECT_CHANGE_DEBOUNCE_DELAY));
     eventManager.listen(player, player.Event.FIRST_PLAY, () => this._onWindowResize());
     this._onWindowResize();
+    eventManager.listen(player, FakeEvent.Type.UI_CLICKED, () => this._startHoverTimeout());
```
yairans, line 229: `UI_CLICKED` is fired from the shell so no need the event listener here at all. just call `_startHoverTimeout` where you fire it

yairans (approved):

yairans:
@JonathanTGold please update the PR desc

PR #638, opened 2021-09-11
fix(FEC-11520): Multi dropdowns are openable in cvaa overlay

JonathanTGold:

issue: the stop propagation (on dropdown click event) prevent the close menus logic handler (which trigerd by docoment click listener) to be trigerd.

fix: remove the soop propagation and move the toggle dropdoen to be happen in the nextt event loop.

sovles: FEC-11520

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/dropdown/dropdown.js
+++ b/src/components/dropdown/dropdown.js
@@ -51,6 +51,10 @@ class DropDown extends Component {
     }
   }
 
+  componentWillUnmount(): void {
+    console.error('');
+  }
+
   /**
    * is given option selected
    *
@@ -82,8 +86,8 @@ class DropDown extends Component {
    * @memberof DropDown
    */
   onClick = (e: Event): void => {
-    e.stopPropagation();
     this.toggleDropDown();
+    setTimeout(() => this.toggleDropDown(), 0);
   };
 
   /**


src/components/dropdown/dropdown.js
```diff
@@ -77,13 +77,12 @@ class DropDown extends Component {
...
    *
-   * @param {Event} e - keyboard event
    * @returns {void}
    * @memberof DropDown
    */
-  onClick = (e: Event): void => {
-    e.stopPropagation();
+  onClick = (): void => {
     this.toggleDropDown();
+    setTimeout(() => this.toggleDropDown(), 0);
```
yairans, line 85: so need to remove from the previous line no?

src/components/dropdown/dropdown.js
```diff
@@ -77,13 +77,12 @@ class DropDown extends Component {
   /**
    * on click handler
    *
-   * @param {Event} e - keyboard event
    * @returns {void}
    * @memberof DropDown
    */
-  onClick = (e: Event): void => {
-    e.stopPropagation();
```
OrenMe, line 85: what are the side effects of removing the stopPropagation? we have additional click handlers that we wouldn't want  them to catch the event.
As only drop down is using the menu maybe we can lift the event listener up to dropdown and expose the handler as a callback prop?

OrenMe: @JonathanTGold ?

JonathanTGold: i did lift it up and now the click outside event is managed from the dropdwon

PR #642, opened 2021-11-21
fix(FEC-11707): V3 - cast on multiple players - when casting started by some player, "Cast" buttons become enabled on all players

SivanA-Kaltura:

Change cast button styles so that cast button would be highlighted only in actively casting player.
Fixes FEC-11707.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/cast/_cast.scss
+++ b/src/components/cast/_cast.scss
@@ -8,12 +8,21 @@
     background-color: rgba(255, 255, 255, 0);
     border: none;
     cursor: pointer;
-    --connected-color: rgba(1, 172, 205, 0.8);
     --disconnected-color: #{$grayscale4};
+    --connected-color: #{$grayscale4};
+
+    &:hover {
+      --disconnected-color: #{$white};
+      --connected-color: #{$white};
+    }
   }
 
-  .cast-button:hover {
-    --disconnected-color: #{$white};
-    --connected-color: #{$brand-color};
+  .cast-button-active {
+    @extend .cast-button;
+    --connected-color: rgba(1, 172, 205, 0.8);
+
+    &:hover {
+      --connected-color: #{$brand-color};
+    }
   }
 }
--- a/src/components/cast/_cast.scss
+++ b/src/components/cast/_cast.scss
@@ -8,21 +8,22 @@
     background-color: rgba(255, 255, 255, 0);
     border: none;
     cursor: pointer;
-    --disconnected-color: #{$grayscale4};
     --connected-color: #{$grayscale4};
+    --disconnected-color: #{$grayscale4};
 
     &:hover {
-      --disconnected-color: #{$white};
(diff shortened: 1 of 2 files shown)


src/components/cast/_cast.scss
```diff
@@ -8,12 +8,22 @@
...
+    &:hover {
+      --connected-color: #{$white};
+      --disconnected-color: #{$white};
+    }
   }
 
-  .cast-button:hover {
-    --disconnected-color: #{$white};
-    --connected-color: #{$brand-color};
+  .cast-button.cast-button-active {
```
yairans, line 20: can move this to the `.player` block? and just use `.cast-button-active` ?

src/components/cast/_cast.scss
```diff
@@ -8,12 +8,22 @@
...
+    --connected-color: rgba(1, 172, 205, 0.8);
+    --disconnected-color: #{$grayscale4};
```
yairans, line 22: why is needed? same for hover

src/components/cast/cast.js
```diff
@@ -73,6 +73,7 @@ class Cast extends Component {
    */
   render(props: any): ?React$Element<any> {
     if (props.isCasting || props.isCastAvailable) {
+      const className = props.isCasting ? `${style.castButton} ${style.castButtonActive}` : style.castButton;
```
yairans, line 76:
```suggestion
      const className = props.isCasting ? [style.castButton, style.castButtonActive].join(' ') : style.castButton;
```
we used to write it like this

SivanA-Kaltura: I think template string is more readable than using join()

src/components/cast/_cast.scss
SivanA-Kaltura: I can use &.cast-button-active inside the cast-button block, is that what you mean ?
wanted to separate because it seemed more readable but not critical

src/components/cast/_cast.scss
SivanA-Kaltura: why is what needed

yairans: `--disconnected-color: #{$grayscale4};` already set

src/components/cast/_cast.scss
yairans: ok

src/components/cast/cast.js
yairans: better to be aligned but not critical

src/components/cast/_cast.scss
SivanA-Kaltura: oh, you're right

src/components/cast/cast.js
SivanA-Kaltura: I care more about readability :)

yairans (approved):

PR #647, opened 2022-01-12
fix(FEC-11733): Hide mouse cursor when it's idle in fullscreen

JonathanTGold:

use the same technique like we use to hide the bottom bar and use the css classes representing the player state to toggle the cursor (hide and show it along with the bottom bar)

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/styles/_shell.scss
+++ b/src/styles/_shell.scss
@@ -11,6 +11,7 @@
   -ms-user-select: none;
   user-select: none;
   -webkit-tap-highlight-color: transparent;
+  cursor: none;
 
   .playback-gui-wrapper,
   .ad-gui-wrapper {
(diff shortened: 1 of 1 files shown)


src/styles/_shell.scss
```diff
@@ -11,6 +11,7 @@
   -ms-user-select: none;
   user-select: none;
   -webkit-tap-highlight-color: transparent;
+  cursor: none;
```
yairans, line 14: 1. there is `.fullscreen` class on the `.player` so I would use it.
2. per the ticket - "We need to monitor its movement and if it’s idle for more than 3 seconds - automatically hide it."

yairans (changes requested):

src/styles/_shell.scss
JonathanTGold: So according to what was agreed with Ron, this will happen in any screen size and will disappear along with the bottom bar

yairans (approved):

PR #682, opened 2022-07-14
fix(FEC-12373): Web][UI][Chrome][Safari] Seek buttons don't spin after clicking on it for the second time

SivanA-Kaltura:

Before https://github.com/kaltura/playkit-js-ui/commit/3b22ffc641969acb3fbef2e97d6a71885630dfa2, when with-animation's componentDidMount was called, the child ref was already set, but the change in the timing of setIsLive caused it to now be null at that point, so it's not possible to register an event listener for child events.

To avoid the issue, I've moved with-animation's event handling into animate, where the child ref is already set.

Resolves FEC-12373

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/utils/with-animation.js
+++ b/src/utils/with-animation.js
@@ -12,36 +12,16 @@ export const withAnimation: Function = (cssClass: string) => (WrappedComponent: 
       ref = createRef();
 
       /**
-       * When component is mounted create event manager instance.
-       * @returns {void}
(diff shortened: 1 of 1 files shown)


giladna (approved):

giladna:
@SivanA-Kaltura  please merge master to here before merge

PR #685, opened 2022-08-10
fix(FEC-12335): Side Panel Item gets refreshed in every 'change media'

JonathanTGold:

**issue:** the engine connector is listening to PLAYER.RESET event (which is called on change media)
and switch from 'playback' preset to 'idle' preset which is a totally different UI
**fix:** cancel the switch and remains on the active preset on change media
 (i think the 'idle' preset is intended only for player start up)

solves: FEC-12335

related prs:
https://github.com/kaltura/kaltura-player-js/pull/562
https://github.com/kaltura/playkit-js/pull/659

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/engine-connector/engine-connector.js
+++ b/src/components/engine-connector/engine-connector.js
@@ -36,7 +36,7 @@ class EngineConnector extends Component {
 
     eventManager.listen(player, player.Event.PLAYER_RESET, () => {
       this.props.updateCurrentTime(0);
-      this.props.updateIsIdle(true);
+      // this.props.updateIsIdle(true);
       this.props.updateIsPlaybackStarted(false);
       this.props.updateDataLoadingStatus(false);
     });
--- a/src/components/engine-connector/engine-connector.js
+++ b/src/components/engine-connector/engine-connector.js
@@ -34,9 +34,11 @@ class EngineConnector extends Component {
     const TrackType = player.Track;
     this.props.updatePrePlayback(!player.config.playback.autoplay);
 
-    eventManager.listen(player, player.Event.PLAYER_RESET, () => {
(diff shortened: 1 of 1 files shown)


SivanA-Kaltura:
@JonathanTGold from what I understand isIdle means that the entry is loading. It's set to false when change source succeeds or fails.

- Will this affect anything else except for the side panel ? What causes the refresh ? Can we have a fix on the side panel side ? 
- It looks like there was a change made specifically to *allow* side panels to be synched (https://github.com/kaltura/playkit-js-ui/commit/c5ea46ac8e86ec532ebf978f654c79a3c7caebb9), so we should be very certain before we remove it.

src/components/engine-connector/engine-connector.js
```diff
@@ -36,7 +36,7 @@ class EngineConnector extends Component {
 
     eventManager.listen(player, player.Event.PLAYER_RESET, () => {
       this.props.updateCurrentTime(0);
-      this.props.updateIsIdle(true);
+      // this.props.updateIsIdle(true);
```
SivanA-Kaltura, line 39: see my other comment

SivanA-Kaltura (changes requested):

src/components/engine-connector/engine-connector.js
JonathanTGold: all the required changes in the side panel side have already been made 

As detailed in the description, the issue is caused by the playkit-ui and not the plugin side because of the transition between the presets

and regarding that https://github.com/kaltura/playkit-js-ui/pull/574

SivanA-Kaltura (approved):

PR #703, opened 2022-11-22
feat(FEC-12700): Make the player colors configurable from studio

JonathanTGold:

1. Organizing the player's palette (and add some new colors) 
2.  Moved all configurable colors from a static (sass) variable to dynamic (css) variable
3.  Added logic to calculate the secondary colors from the primary color (so that a user can only choose the primary color in the studio and the secondary ones will be derived from the primary one)
4. move the date-url of  dynamic SVG icons (Those whose color changes in active mode to the color of the main color) to a variable so they can be changed at runtime (our current icons infrastructures uses sass functions to generate the data-url which are not useful at runtime but only at compile time)
5. add theme manager service that get the user configured style and apply the style

related pr: https://github.com/kaltura/playkit-js-ui-managers/pull/18

solves: FEC-12700

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/icon/icon.scss
+++ b/src/components/icon/icon.scss
@@ -249,7 +249,8 @@
 }
 
 .icon-vr-stereo-full {
-  background-image: icon(vrStereoFull, $brand-color);
+  --playkit-icon-data-url: #{icon(vrStereoFull, $brand-color)};
+  background-image: var(--playkit-icon-data-url);
 }
 
 .icon-chromecast {
@@ -257,7 +258,8 @@
 }
 
 .icon-chromecast-brand {
-  background-image: icon(chromecast, $brand-color);
+  --playkit-icon-data-url: #{icon(chromecast, $brand-color)};
+  background-image: var(--playkit-icon-data-url);
 }
 
 .icon-next {
@@ -283,8 +285,9 @@
 }
 
 .icon-quality-hd-active {
+  --playkit-icon-data-url: #{icon(qualityHdActive, $brand-color, '16', '0 0 16 16')};
   &:after {
-    background-image: icon(qualityHdActive, $brand-color, '16', '0 0 16 16');
+    background-image: var(--playkit-icon-data-url);
   }
 }
 
@@ -295,8 +298,9 @@
 }
 
 .icon-quality-4k-active {
+  --playkit-icon-data-url: #{icon(quality4kActive, $brand-color, '16', '0 0 16 16')};
   &:after {
-    background-image: icon(quality4kActive, $brand-color, '16', '0 0 16 16');
+    background-image: var(--playkit-icon-data-url);
   }
 }
 
@@ -307,8 +311,9 @@
 }
 
 .icon-quality-8k-active {
+  --playkit-icon-data-url: #{icon(quality8kActive, $brand-color, '16', '0 0 16 16')};
   &:after {
-    background-image: icon(quality8kActive, $brand-color, '16', '0 0 16 16');
+    background-image: var(--playkit-icon-data-url);
   }
 }
 
--- a/src/components/icon/icon.scss
+++ b/src/components/icon/icon.scss
@@ -1,5 +1,3 @@
-@import '~styles/variables';
-
 @function svg-url($svg, $size, $viewBox) {
   $svg: '<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="#{$viewBox}" width="#{$size}" height="#{$size}">#{$svg}</svg>';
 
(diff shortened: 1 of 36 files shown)


SivanA-Kaltura:
@JonathanTGold how would a plugin access the new values now ? what refactor should be made if a plugin was accessing the values through the js object ? (KalturaPlayer.ui.style)

SivanA-Kaltura (changes requested):

semarche-kaltura:
> @JonathanTGold how would a plugin access the new values now ? what refactor should be made if a plugin was accessing the values through the js object ? (KalturaPlayer.ui.style)

V7 plugins uses new style variables (changes already implemented, waiting till current PR got merged and deployed). No issues found with using new styles variables.

semarche-kaltura (approved):

src/components/icon/icon.scss
```diff
@@ -249,15 +249,17 @@
 }
 
 .icon-vr-stereo-full {
-  background-image: icon(vrStereoFull, $brand-color);
+  --playkit-icon-data-url: #{icon(vrStereoFull, $brand-color)};
```
SivanA-Kaltura, line 252: where do the default values for the variables come from ?

SivanA-Kaltura (changes requested):

src/components/icon/icon.scss
JonathanTGold: sass icon function

src/utils/themes-manager.js
```diff
@@ -0,0 +1,79 @@
...
+const dynamicColoredIconsSvgUrlVars = [
+  '--playkit-icon-check-active-url',
+  '--playkit-icon-data-url',
+  '--playkit-icon-chromecast-url',
+  '--playkit-icon-quality-HD-active-url',
+  '--playkit-icon-quality-4K-active-url',
+  '--playkit-icon-quality-8K-active-url'
+];
+
+// eslint-disable-next-line require-jsdoc
```
SivanA-Kaltura, line 27: you can disable it for the entire file at once.
do you want to add this in a separate task ?

src/utils/color-format-convertors.js
```diff
@@ -0,0 +1,73 @@
...
+  } else if (r === max) {
+    h = (g - b) / delta;
+  } else if (g === max) {
+    h = 2 + (b - r) / delta;
+  } else if (b === max) {
+    h = 4 + (r - g) / delta;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
SivanA-Kaltura, line 50: do you need @ts-ignore for playkit-ui ?

src/utils/themes-manager.js
```diff
@@ -0,0 +1,79 @@
...
+
+  // eslint-disable-next-line require-jsdoc
+  constructor(player: Object, config: UserTheme) {
+    this.player = player;
+    this.config = config;
+  }
+
+  // eslint-disable-next-line require-jsdoc
+  applyUserTheme(): void {
+    if (this.config) {
```
SivanA-Kaltura, line 40: can this be not set ?

src/utils/themes-manager.js
```diff
@@ -0,0 +1,79 @@
...
+  // eslint-disable-next-line require-jsdoc
+  getHueComponent(hex: string): number {
+    const hsl = hexToHsl(hex);
+    return hsl[0];
+  }
+
+  // eslint-disable-next-line require-jsdoc
+  setColors(config: ThemesManagerConfig): void {
+    if (config.colors.primary) {
+      // this.setColor(cssVarNames.colors.primary, config.colors.primary);
```
SivanA-Kaltura, line 54: remove ?

src/utils/themes-manager.js
```diff
@@ -0,0 +1,79 @@
...
+
+    for (const color in config.colors) {
+      this.setColor(cssVarNames.colors[color], config.colors[color]);
+    }
+  }
+
+  // eslint-disable-next-line require-jsdoc
+  setColor(cssVarName: string, color: string): void {
+    const hue = this.getHueComponent(color);
+    document.querySelector<HTMLElement>('.playkit-player').style.setProperty(cssVarName, `${hue}deg`);
```
SivanA-Kaltura, line 66: this query will return every player on the page, if theres more than 1. meaning you'll also be setting the style for every player on the page.

SivanA-Kaltura (changes requested):

src/utils/color-format-convertors.js
JonathanTGold: fixed

src/utils/themes-manager.js
JonathanTGold: yes it does, the default colors comes from the css files and bundled with the player , the config just override

src/utils/themes-manager.js
JonathanTGold: document.querySelector returns only one element (the first within the document that matches the specified selector) but anyway the issue is still there that it will return the wrong player so what do you suggest how to identify the specific player ?

src/utils/themes-manager.js
JonathanTGold: fixed

src/utils/color-format-convertors.js
JonathanTGold: fixed

ErezKaltura (dismissed):

src/utils/themes-manager.js
SivanA-Kaltura: you can use something like this for example
https://github.com/kaltura/playkit-js-visibility/blob/1e125cbc275e95918a7cc9ac5b31d5501baa5cd4/src/visibility.js#L96

SivanA-Kaltura (approved):

PR #704, opened 2022-11-27
fix(FEC-12763): Volume slider is inaccessible for JAWS screen reader

Tzipi-kaltura:

**the issue:**
JAWS and NVDA screen readers wasn't able to open the volume slider by up/down arrow

**the root cause:**
JWAS and NVDA have default behavior for the up/down keyboard events. so it doesn't recognize our handling 

**the solution:**
add role='application' to enable non-default behavior for keyboard events

solves FEC-12763

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/volume/volume.js
+++ b/src/components/volume/volume.js
@@ -400,6 +400,7 @@ class Volume extends Component {
 
     return (
       <ButtonControl
+        role="application"
         name={COMPONENT_NAME}
         ref={c => (c ? (this._volumeControlElement = c) : undefined)}
         className={controlButtonClasses}


JonathanTGold (changes requested):
...In this mode, the web author is completely responsible for handling any and all keyboard input, focus management, and other interactions and cannot assume assistive technologies would do any processing on their end.

[see here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/application_role)

this disables the Native browser's behaviour  of this element and all of its decedents

PR #741, opened 2023-03-13
fix(FEC-12910): allow play/pause using spacebar/enter when focused on player or seekbar

lianbenjamin:

allow play/pause using spacebar/enter while focusing on seekbar component or player.

Solves FEC-12910

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/shell/shell.js
+++ b/src/components/shell/shell.js
@@ -81,7 +81,7 @@ class Shell extends Component {
   hoverTimeout: ?TimeoutID;
   _environmentClasses: Array<string>;
   _playerResizeWatcher: ResizeWatcher;
-  _playerRef: HTMLDivElement;
+  _playerRef: ?HTMLDivElement;
 
   /**
    * on mouse over, add hover class (shows the player ui) and timeout of 3 seconds bt default or what pass as prop configuration to component


giladna (approved):

giladna:
@lianbenjamin  please fix the flow check

src/components/seekbar/seekbar.js
```diff
@@ -302,9 +302,24 @@ class SeekBar extends Component {
...
   };
 
+  /**
+   * toggle play/pause
+   *
+   * @returns {void}
+   * @memberof SeekBar
+   */
+  togglePlayPause = (): void => {
+    this.props.player.paused ? this.props.player.play() : this.props.player.pause();
```
SivanA-Kaltura, line 320: i think that the player is paused

src/components/shell/shell.js
```diff
@@ -191,6 +192,10 @@ class Shell extends Component {
 
     if (this.state.nav && (e.keyCode === KeyMap.ENTER || e.keyCode === KeyMap.SPACE)) {
       this.unMuteFallback();
+      if (e.srcElement.contains(this._playerRef)) {
```
SivanA-Kaltura, line 195: why is the condition needed ?

SivanA-Kaltura (changes requested):

src/components/shell/shell.js
lianbenjamin: @SivanA-Kaltura  it seems that if you use the keyboard for other controls on the player it bubbles to the parent (the root of player), so e want to make sure we toggle play/pause when the event came from the video player element, as we don't want to do it if for example the user pressed the spacebar for seek forward button

SivanA-Kaltura: won't this catch every spacebar click on the root document as well though ?

SivanA-Kaltura (changes requested):

src/components/shell/shell.js
lianbenjamin: no, I checked and verified; the `event.srcElement` is different for the other controls..

semarche-kaltura (approved):

SivanA-Kaltura (approved):

PR #752, opened 2023-04-09
feat(FEC-13040): Allow player settings menu ("Gear") to be extended & modify

JonathanTGold:

solves FEC-1304

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/flow-typed/types/ui-options.js
+++ b/flow-typed/types/ui-options.js
@@ -12,5 +12,6 @@ declare type UIOptionsObject = {
   uiComponents?: Array<KPUIComponent>,
   translations?: {[langKey: string]: Object},
   locale?: string;
-  userTheme?: UserTheme
+  userTheme?: UserTheme,
+  userSettingsMenus: any[],
 };
--- a/src/components/settings/settings.js
+++ b/src/components/settings/settings.js
@@ -36,7 +36,8 @@ const mapStateToProps = state => ({
   videoTracks: state.engine.videoTracks,
   isMobile: state.shell.isMobile,
   isSmallSize: state.shell.isSmallSize,
-  isLive: state.engine.isLive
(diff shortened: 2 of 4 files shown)


flow-typed/types/ui-options.js
```diff
@@ -12,5 +12,6 @@ declare type UIOptionsObject = {
   uiComponents?: Array<KPUIComponent>,
   translations?: {[langKey: string]: Object},
   locale?: string;
-  userTheme?: UserTheme
+  userTheme?: UserTheme,
+  userSettingsMenus: any[],
```
SivanA-Kaltura, line 16: if userSettingsMenus always has the same structure (icon, label, options, onMenuChosen) make a type for it.

src/components/settings/settings.js
```diff
@@ -375,6 +376,9 @@ class Settings extends Component {
             ) : (
               <SmartContainerItem icon="speed" label={props.speedLabelText} options={speedOptions} onMenuChosen={this.onSpeedChange} />
             )}
+            {props.userSettingsMenus.map((menu, index) => (
+              <SmartContainerItem key={index} icon={menu.icon} label={menu.label} options={menu.options} onMenuChosen={menu.onMenuChosen} />
```
SivanA-Kaltura, line 380: what happens if one of the fields of menu is not set (e.g. options or label are missing) ? if some fields are optional - there should probably be a default for them, otherwise you should ignore a menu without values in all of the fields

SivanA-Kaltura (changes requested):

flow-typed/types/ui-options.js
JonathanTGold: removed- refactored

SivanA-Kaltura (approved):

PR #758, opened 2023-05-01
fix(SUP-37036): Videos not conserving the player customization

Tzipi-kaltura:
**issue:**
In the case of several players on a page, the first player is always colored and **only** the first player is colored.

**solution:**
select the player div element from the dom by the targetId and then select the another elements from the player and not directly from the dom.

solves [SUP-37036](https://kaltura.atlassian.net/browse/SUP-37036)

### CheckLists

- [ ] Docs have been updated

[SUP-37036]: https://kaltura.atlassian.net/browse/SUP-37036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/themes-manager.js
+++ b/src/utils/themes-manager.js
@@ -30,11 +30,13 @@ const dynamicColoredIconsSvgUrlVars = [
 export class ThemesManager {
   player: Object;
   config: Object;
+  playerElement: HTMLElement;
 
   // eslint-disable-next-line require-jsdoc
-  constructor(player: Object, config: ?UserTheme) {
+  constructor(player: Object, config: ?UserTheme, targetId: string) {
     this.player = player;
     this.config = config;
+    this.playerElement = document.querySelector(`#${targetId}`);
   }
 
   /**
@@ -74,10 +76,14 @@ export class ThemesManager {
    */
   setAccentOrAcknowledgementColor(colorTitle: string, color: string): void {
     const [hue, saturation, lightness] = hexToHsl(color);
-    document.querySelector(`.${style.player}`)?.style.setProperty(ACTUAL_USED_CSS_VAR.replace('{name}', colorTitle), color);
-    document.querySelector(`.${style.player}`)?.style.setProperty(HSL_HUE_CSS_VAR.replace('{name}', colorTitle), `${Math.round(hue)}deg`);
-    document.querySelector(`.${style.player}`)?.style.setProperty(HSL_SATURATION_CSS_VAR.replace('{name}', colorTitle), `${Math.round(saturation)}%`);
-    document.querySelector(`.${style.player}`)?.style.setProperty(HSL_LIGHTNESS_CSS_VAR.replace('{name}', colorTitle), `${Math.round(lightness)}%`);
+    this.playerElement.querySelector(`.${style.player}`)?.style.setProperty(ACTUAL_USED_CSS_VAR.replace('{name}', colorTitle), color);
+    this.playerElement.querySelector(`.${style.player}`)?.style.setProperty(HSL_HUE_CSS_VAR.replace('{name}', colorTitle), `${Math.round(hue)}deg`);
+    this.playerElement
+      .querySelector(`.${style.player}`)
+      ?.style.setProperty(HSL_SATURATION_CSS_VAR.replace('{name}', colorTitle), `${Math.round(saturation)}%`);
+    this.playerElement
+      .querySelector(`.${style.player}`)
+      ?.style.setProperty(HSL_LIGHTNESS_CSS_VAR.replace('{name}', colorTitle), `${Math.round(lightness)}%`);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/utils/themes-manager.js
```diff
@@ -74,10 +76,14 @@ export class ThemesManager {
    */
   setAccentOrAcknowledgementColor(colorTitle: string, color: string): void {
     const [hue, saturation, lightness] = hexToHsl(color);
-    document.querySelector(`.${style.player}`)?.style.setProperty(ACTUAL_USED_CSS_VAR.replace('{name}', colorTitle), color);
-    document.querySelector(`.${style.player}`)?.style.setProperty(HSL_HUE_CSS_VAR.replace('{name}', colorTitle), `${Math.round(hue)}deg`);
-    document.querySelector(`.${style.player}`)?.style.setProperty(HSL_SATURATION_CSS_VAR.replace('{name}', colorTitle), `${Math.round(saturation)}%`);
-    document.querySelector(`.${style.player}`)?.style.setProperty(HSL_LIGHTNESS_CSS_VAR.replace('{name}', colorTitle), `${Math.round(lightness)}%`);
+    this.playerElement.querySelector(`.${style.player}`)?.style.setProperty(ACTUAL_USED_CSS_VAR.replace('{name}', colorTitle), color);
```
lianbenjamin, line 79: could be a case where the playerElement is null or undefined? for example, if it can't get the element by the provided targetId. if that's a case we need to handle, then you'd need to add `?` in all places where you try to access `playerElement`.

lianbenjamin (changes requested):

src/utils/themes-manager.js
```diff
@@ -30,11 +30,13 @@ const dynamicColoredIconsSvgUrlVars = [
...
   player: Object;
   config: Object;
+  playerElement: HTMLElement;
 
   // eslint-disable-next-line require-jsdoc
-  constructor(player: Object, config: ?UserTheme) {
+  constructor(player: Object, config: ?UserTheme, targetId: string) {
     this.player = player;
     this.config = config;
+    this.playerElement = document.querySelector(`#${targetId}`);
```
JonathanTGold, line 39:
```suggestion
    this.playerContainerElement = document.querySelector(`#${targetId}`);
```

JonathanTGold (changes requested):

src/utils/themes-manager.js
Tzipi-kaltura: done

src/utils/themes-manager.js
Tzipi-kaltura: done

src/utils/themes-manager.js
```diff
@@ -74,10 +76,16 @@ export class ThemesManager {
    */
   setAccentOrAcknowledgementColor(colorTitle: string, color: string): void {
     const [hue, saturation, lightness] = hexToHsl(color);
-    document.querySelector(`.${style.player}`)?.style.setProperty(ACTUAL_USED_CSS_VAR.replace('{name}', colorTitle), color);
-    document.querySelector(`.${style.player}`)?.style.setProperty(HSL_HUE_CSS_VAR.replace('{name}', colorTitle), `${Math.round(hue)}deg`);
-    document.querySelector(`.${style.player}`)?.style.setProperty(HSL_SATURATION_CSS_VAR.replace('{name}', colorTitle), `${Math.round(saturation)}%`);
-    document.querySelector(`.${style.player}`)?.style.setProperty(HSL_LIGHTNESS_CSS_VAR.replace('{name}', colorTitle), `${Math.round(lightness)}%`);
+    this.playerContainerElement?.querySelector(`.${style.player}`)?.style.setProperty(ACTUAL_USED_CSS_VAR.replace('{name}', colorTitle), color);
```
JonathanTGold, line 79: Adding validation without referring to the invalidity case could make it difficult to debug 
I would warp it all with an if-else statement and log a warning like 'the player id: ${id} was not found, user custom theme has not been applied...'
and then you can remove the **?** from inside the if

JonathanTGold (changes requested):

src/utils/themes-manager.js
JonathanTGold: target id comes from the config, so if the target is wrong the player was not up and running in the first place

lianbenjamin: @JonathanTGold  right, thanks for pointing that out.
@Tzipi-kaltura so you can remove the `?`

src/utils/themes-manager.js
```diff
@@ -30,11 +30,13 @@ const dynamicColoredIconsSvgUrlVars = [
 export class ThemesManager {
   player: Object;
   config: Object;
+  playerContainerElement: HTMLElement | null;
```
JonathanTGold, line 33:
```suggestion
  playerContainerElement!: HTMLElement;
```
after you add this operator you will be able to get rid of the **?** operator below

JonathanTGold (changes requested):

JonathanTGold (approved):

lianbenjamin (approved):

PR #785, opened 2023-09-10
fix(ADA-31): [V7-Acc] advanced captions modal heading marking correction

Tzipi-kaltura:

**issue:**
in advanced captions modal, when open the modal, screen reader reads only : "dialog".

**solution:**
adding aria-labeledby according to the title of the modal so screen reader announce: "Advanced captions settings dialog" 

solves [ADA-31](https://kaltura.atlassian.net/browse/ADA-31)

### CheckLists

- [ ] Docs have been updated

[ADA-31]: https://kaltura.atlassian.net/browse/ADA-31?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/cvaa-overlay/cvaa-overlay.js
+++ b/src/components/cvaa-overlay/cvaa-overlay.js
@@ -11,6 +11,7 @@ import {withEventDispatcher} from 'components/event-dispatcher';
 import {withLogger} from 'components/logger';
 import {MainCaptionsWindow} from 'components/cvaa-overlay/main-captions_window';
 import {CustomCaptionsWindow} from 'components/cvaa-overlay/custom-captions-window';
+import {withText} from 'preact-i18n';
 
 /**
  * mapping state to props
@@ -41,6 +42,7 @@ const COMPONENT_NAME = 'CVAAOverlay';
 @withLogger(COMPONENT_NAME)
 @withEventDispatcher(COMPONENT_NAME)
 @withKeyboardA11y
+@withText({cvvaDialogText: 'cvaa.title'})
 class CVAAOverlay extends Component {
   /**
    * componentWillUnmount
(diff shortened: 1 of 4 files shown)


JonathanTGold (changes requested):
the overlay is generic component , so you can’t rely on the fact that the parent component will have the same id, 
my suggestion is to add label prop to the dialog component with Default = 'dialog'   and then it will have to be supplied from outside , that way the generic is not breaken

src/components/smart-container/smart-container.js
```diff
@@ -94,7 +96,12 @@ class SmartContainer extends Component {
     props.clearAccessibleChildren();
     return this.isPortal ? (
       createPortal(
-        <Overlay open onClose={props.onClose} handleKeyDown={this.props.handleKeyDown} addAccessibleChild={this.props.addAccessibleChild}>
+        <Overlay
+          open
+          onClose={props.onClose}
+          handleKeyDown={this.props.handleKeyDown}
+          addAccessibleChild={this.props.addAccessibleChild}
+          label={props.settingsText}>
```
JonathanTGold, line 104: add {props.settingsText + " " + 'dialog'}, otherwise the screenreader won’t indicate it is a dialog

src/components/cvaa-overlay/cvaa-overlay.js
```diff
@@ -132,7 +134,13 @@ class CVAAOverlay extends Component {
...
     props.clearAccessibleChildren();
     return (
-      <Overlay open handleKeyDown={this.props.handleKeyDown} addAccessibleChild={this.props.addAccessibleChild} onClose={props.onClose} type="cvaa">
+      <Overlay
+        open
+        handleKeyDown={this.props.handleKeyDown}
+        addAccessibleChild={this.props.addAccessibleChild}
+        onClose={props.onClose}
+        type="cvaa"
+        label={props.cvvaDialogText}>
```
JonathanTGold, line 143: add `{props.settingsText + " " + 'dialog'}`, otherwise the screenreader won’t indicate it is a dialog

"Perhaps we should consider adding translations for this. You can ask Dana or David for assistance. In the meantime, I suppose it's acceptable as it is."

JonathanTGold (changes requested):

src/components/smart-container/smart-container.js
Tzipi-kaltura: screen reader already reads "dialog" because of the role="dialog"

PR #786, opened 2023-09-11
fix(ADA-27):[V7-Acc] Enhance buttons labels description

inbalvasserman:
issue:
in seek backwards and seek forward user don't know how much it will seek

solution:
change label to seek 10 seconds backward/forward


JonathanTGold (changes requested):
it is not always 10 sec, it is configurable and subjected to changes,
my suggestion is to add another field field rewind10 and  forward10 , get the translations for them , and then dynamically  choose the appropriate translation according the configuraion

src/components/forward/forward.js
```diff
@@ -30,6 +30,21 @@ const mapStateToProps = state => ({
...
+ * translates
+ * @param {any} props - Props
+ * @returns {Object} - The object translations
+ */
+const translates = (props: any) => ({
+  forwardText: !props.step ? (
+    <Text id={'controls.forward'}>Seek forward</Text>
+  ) : (
+    <Text id={'controls.secondsForward'} fields={{seconds: props.step}}>
+      {`Seek ${props.step} seconds forward`}
```
JonathanTGold, line 44: Why are you repeating the same string that you wrote in the json file ????   

It makes no sense and redundant 
See here : https://www.npmjs.com/package/preact-i18n#pluralization-and-templating

translations/en.i18n.json
```diff
@@ -8,7 +8,9 @@
       "fullscreen": "Fullscreen",
       "fullscreenExit": "Exit fullscreen",
       "rewind": "Seek backwards",
+      "secondsRewind": "Seek {{seconds}} seconds backwards",
```
SivanA-Kaltura, line 11: do we have a ticket for adding translations in other languages ?

SivanA-Kaltura (changes requested):

translations/en.i18n.json
inbalvasserman: yes, created FEC-13468

JonathanTGold (approved):

SivanA-Kaltura (approved):

PR #800, opened 2023-10-01
fix(FEC-13377): Player v7| audio player entry info| Entry info does not change when changing audio entries

SivanA-Kaltura:

Audio player details component does not re-render on entry change, probably due to its location in the component hierarchy. 
To force the component to re-render - added a listener to source change events.

Resolves FEC-13377

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/audio-entry-details/audio-entry-details.js
+++ b/src/components/audio-entry-details/audio-entry-details.js
@@ -12,10 +12,14 @@ import {withPlayer} from '../player';
 import {ExpandableText} from '../expandable-text';
 import {Scrollable} from '../scrollable';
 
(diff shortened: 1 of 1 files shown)


JonathanTGold (changes requested):
you just need to add shouldUpdate={true} prop to PlayerArea component in playback.js and the problem should be solved 

your listener and refactor is redundant

JonathanTGold (approved):

SivanA-Kaltura:
@JonathanTGold As we've spoken earlier, it looks like player state changes too quickly and does not trigger the re-render of the component, so using an event listener is an ugly but working solution

semarche-kaltura (approved):

PR #823, opened 2023-12-11
fix(FEC-13518): timeline blue bar is on top of button tooltips

lianbenjamin:

- bug fix.

This issue is a regression from this [PR](https://github.com/kaltura/playkit-js-ui/pull/782/files).

**the issue:**
Since the z-index of the seekbar area was increased, it is now overlaying the tooltip of the control buttons which is inside a sibling container of the bottom bar. it also overlays for example the hover container of playlist (next and prev items).

**solution:**
matching the z-index of controls-container to the seekbar.

Solves FEC-13518

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/components/tooltip/_tooltip.scss
+++ b/src/components/tooltip/_tooltip.scss
@@ -9,7 +9,6 @@
   display: inline-block;
   height: 100%;
   width: 100%;
-  z-index: 1;
   .tooltip-label {
     visibility: hidden;
     background-color: $tooltip-background-color;
--- a/src/components/tooltip/_tooltip.scss
+++ b/src/components/tooltip/_tooltip.scss
@@ -9,6 +9,7 @@
   display: inline-block;
   height: 100%;
   width: 100%;
+  z-index: 1;
   .tooltip-label {
     visibility: hidden;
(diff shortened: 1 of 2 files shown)


src/components/tooltip/_tooltip.scss
```diff
@@ -9,6 +9,7 @@
   display: inline-block;
   height: 100%;
   width: 100%;
+  z-index: 1;
```
SivanA-Kaltura, line 12: don't we have other elements with z-index 1 as well ? is any of them in a position where it can hide seekbar ?
i think overlays have z-index 1 as well

SivanA-Kaltura (changes requested):

src/components/tooltip/_tooltip.scss
lianbenjamin: not sure I understand the question.. this issue is a regression from this [PR](https://github.com/kaltura/playkit-js-ui/pull/782/files). since we increased the z-index of the seekbar area, it overlays the tooltip of the control buttons which is inside a sibling container of the bottom bar.

SivanA-Kaltura (approved):

PR #828, opened 2023-12-26
fix(FEC-13535): add classnames to expandable text

semarche-kaltura:

resolves: https://kaltura.atlassian.net/browse/FEC-13535

### CheckLists

- [ ] Docs have been updated


src/components/expandable-text/expandable-text.js
```diff
@@ -86,7 +86,7 @@ const ExpandableText = withText({
   }
 
   return (
-    <div>
+    <div className={['expandable-text-container', isTextExpanded ? 'text-expanded' : ''].join(' ')}>
```
lianbenjamin, line 89: hey @semarche-kaltura , is it possible to write the class names from styles like we usually do, instead of hard-coding them? or export them to const?

lianbenjamin (changes requested):

src/components/expandable-text/expandable-text.js
```diff
@@ -86,7 +86,7 @@ const ExpandableText = withText({
   }
 
   return (
```
SivanA-Kaltura, line 88: why is this needed ?

SivanA-Kaltura (changes requested):

src/components/expandable-text/expandable-text.js
SivanA-Kaltura, line 89: why is this needed ? this would create a coupling with these texts to somewhere else, and it could break if it's changed in one of the repos

SivanA-Kaltura (changes requested):

SivanA-Kaltura (approved):

PR #844, opened 2024-01-16
feat(FEC-13507): Doc preset and methods

semarche-kaltura:

base PR: https://github.com/kaltura/playkit-js-doc-player/pull/1

solves: https://kaltura.atlassian.net/browse/FEC-13507

**Issue:**

**Fix:**


src/components/volume/volume.tsx
```diff
@@ -123,7 +123,7 @@ class Volume extends Component<any, any> {
    * @returns {boolean} - whether to render the component
    */
   _shouldRender(): boolean {
-    const isActive = !this.props.player.isImage();
+    const isActive = !(this.props.player.isImage() || this.props.player.isDoc());
```
SivanA-Kaltura, line 126: just a reminder - if we change the name to document in providers we should change it in other places as well to be aligned.
if we leave it as doc then its fine.

SivanA-Kaltura (changes requested):

lianbenjamin (approved):

src/reducers/engine.ts
```diff
@@ -40,6 +40,7 @@ export const types = {
   UPDATE_IS_LIVE: `${component}/UPDATE_IS_LIVE`,
   UPDATE_IS_DVR: `${component}/UPDATE_IS_DVR`,
   UPDATE_IS_IMG: `${component}/UPDATE_IS_IMG`,
+  UPDATE_IS_DOC: `${component}/UPDATE_IS_DOC`,
```
SivanA-Kaltura, line 43: not critical, but it would be nice to update this to document as well, because it updates the value of isDocument field

SivanA-Kaltura (changes requested):

src/reducers/engine.ts
semarche-kaltura: I kept it aligned with other reducers keys like `UPDATE_IS_IMG` where source named `image`, etc

SivanA-Kaltura (approved):

PR #775, opened 2023-07-30
fix(FEC-13118):  pause/play playback using keyboard - A11Y issue

JonathanTGold:
### Description of the Changes

solves FEC-13118

**issue:**  the keyboard event listener is listening only to events within the player container element scope 

### CheckLists

- [x] changes have been done against master branch, and PR does not conflict
- [ ] new unit / functional tests have been added (whenever applicable)
- [ ] test are passing in local environment
- [ ] Travis tests are passing (or test results are not worse than on master branch :))
- [ ] Docs have been updated

PR #848, opened 2024-01-25
fix(ADA-190): Focus back on the player from other elements

Tzipi-kaltura:

**Issue:**
There is no option to focus on the player via keyboard when the control bar is hidden

**Fix:**
Listen to tab press on the page, display the control bar and move the focus to the playerContainer or to the last element on the player (according to the user press: tab or shift tab)

#### Resolves [ADA-190](https://kaltura.atlassian.net/browse/ADA-190)

[ADA-190]: https://kaltura.atlassian.net/browse/ADA-190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/shell/shell.tsx
+++ b/src/components/shell/shell.tsx
@@ -11,6 +11,7 @@ import {withEventDispatcher} from '../event-dispatcher';
 import {withLogger} from '../logger';
 import {ResizeWatcher} from '../../utils/resize-watcher';
 import {debounce} from '../../utils/debounce';
+import {focusElement} from '../../utils';
 
 /**
  * mapping state to props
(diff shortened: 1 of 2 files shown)


src/components/shell/shell.tsx
```diff
@@ -344,6 +345,29 @@ class Shell extends Component<any, any> {
     }
   }
 
+  _setFocusOnPlayer(): void {
+    const {eventManager, player} = this.props;
+    eventManager.listen(player, player.Event.FIRST_PLAY, () => {
+      const playerContainer: HTMLDivElement = document.getElementById(this.props.targetId) as HTMLDivElement;
+      playerContainer.setAttribute('tabindex', '0');
```
JonathanTGold, line 352: You should not manipulate the dom directly but set conditional attribute in jsx

JonathanTGold (changes requested):

JonathanTGold (changes requested):
Is not it the same issue ?
https://github.com/kaltura/playkit-js-ui/pull/775

JonathanTGold (changes requested):
https://github.com/kaltura/playkit-js-ui/pull/775

See this PR

Tzipi-kaltura:
> #775
> 
> See this PR

it's a different issue. here we need to focus on the player after control bar is hidden. in your pr it's moving to focus after click play.

JonathanTGold (changes requested):
after through investigation with Tzippi , the flow should be as following:

the root of the issue is that the player doesn’t contain any **tabblable** elements (since bottom bar is removed from the DOM completely on hover state = false)

1. fix difference bottom bar behaviour between mouse click and tab movements (the bottom bar should disappear in tab movement as well)
2. bottom bar should show up (as it does on hover) when focued with tab movements

SivanA-Kaltura:
@Tzipi-kaltura Closing this PR as this issue will be solved by a separate PR

PR #851, opened 2024-01-29
fix(ADA-97): The small play button loses the focus after play

Tzipi-kaltura:

**Issue:**
When click on the small play button in the control bar via keyboard, the focus move to playerContainer instead of stay of the play button.

**Fix:**
Check only if the play was done from the large player move the focus to playerContainer(if user click and this is the first play)

#### Resolves [ADA-97](https://kaltura.atlassian.net/browse/ADA-97)

[ADA-97]: https://kaltura.atlassian.net/browse/ADA-97?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/play-pause/play-pause.tsx
+++ b/src/components/play-pause/play-pause.tsx
@@ -57,8 +57,8 @@ class PlayPause extends Component<any, any> {
   componentDidMount(): void {
     const {eventManager, player} = this.props;
     const playerContainer: HTMLDivElement = document.getElementById(player.config.ui.targetId) as HTMLDivElement;
-    eventManager.listen(player, player.Event.UI.USER_CLICKED_PLAY, () => {
-      eventManager.listen(player, player.Event.Core.FIRST_PLAY, () => {
+    eventManager.listenOnce(player, player.Event.UI.USER_CLICKED_PLAY, () => {
+      eventManager.listenOnce(player, player.Event.Core.FIRST_PLAY, () => {
         playerContainer.focus();
       });
     });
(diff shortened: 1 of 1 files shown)


src/components/play-pause/play-pause.tsx
```diff
@@ -58,7 +58,9 @@ class PlayPause extends Component<any, any> {
     const {eventManager, player} = this.props;
     const playerContainer: HTMLDivElement = document.getElementById(player.config.ui.targetId) as HTMLDivElement;
     eventManager.listen(player, player.Event.UI.USER_CLICKED_PLAY, () => {
-      playerContainer.focus();
+      eventManager.listen(player, player.Event.Core.FIRST_PLAY, () => {
```
SivanA-Kaltura, line 61: what if play is called via player API ?
also, does it matter if it was initiated via button click or mouse click ?

SivanA-Kaltura (changes requested):

src/components/play-pause/play-pause.tsx
Tzipi-kaltura: yes, only for first user's click on the large play button.

SivanA-Kaltura (changes requested):

src/components/play-pause/play-pause.tsx
Tzipi-kaltura: right, changed it to listenOnce

SivanA-Kaltura (approved):

PR #852, opened 2024-02-01
fix(FEC-13680): handle logo link with ima

lianbenjamin:

- **bugfix**

**Issue:**
A use-case where the `medialoaded` event has been fired before logo component was mounted.
in case ima is enabled and the media is being preloaded, listening to `medialoaded` event happens too late (after it was already fired). This is causing not replacing the `{entryId}` in the url, which breaks the url href of the logo.

**Fix:**
before listening to `medialoaded` event, first check if the entry id already exists.

#### Resolves [FEC-13680](https://kaltura.atlassian.net/browse/FEC-13680)

[FEC-13680]: https://kaltura.atlassian.net/browse/FEC-13680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/logo/logo.tsx
+++ b/src/components/logo/logo.tsx
@@ -40,7 +40,7 @@ class Logo extends Component<any, any> {
    */
   constructor(props: any) {
     super(props);
-    this.setState({isUrlClickable: true, urlLink: props.config.url});
+    this.setState({isUrlClickable: typeof props.config.url === 'string', urlLink: props.config.url});
   }
 
   /**
@@ -64,22 +64,31 @@ class Logo extends Component<any, any> {
     const url = this.props.config.url;
     if (url && url.indexOf(ENTRY_VAR) !== -1) {
       const {player, eventManager} = this.props;
-      const entryId = player.sources.id;
-      if (typeof entryId === 'string') {
-        this.setState({urlLink: url.replace(ENTRY_VAR, entryId)});
-      } else {
-        this.setState({isUrlClickable: false});
+      if (!this._maybeSetLogoUrlWithEntryId(url)) {
         eventManager.listen(player, player.Event.MEDIA_LOADED, () => {
-          const entryId = player.sources.id;
-          if (typeof entryId === 'string') {
-            this.setState({urlLink: url.replace(ENTRY_VAR, entryId), isUrlClickable: true});
-          } else {
-            this.props.logger.debug(`Logo url was not replaced; entry id was not found.`);
-            this.setState({isUrlClickable: false});
-          }
+          this._maybeSetLogoUrlWithEntryId(url);
         });
       }
     }
+  }
+
+  /**
+   * sets the url with the entry id
+   * @param {string} url - the url configured on the logo
+   * @returns {boolean} - whether the url was set with entry id
+   * @memberof Logo
+   */
+  private _maybeSetLogoUrlWithEntryId(url: string): boolean {
+    const {player} = this.props;
+    const entryId = player.sources.id;
+    if (typeof entryId === 'string') {
+      this.setState({urlLink: url.replace(ENTRY_VAR, entryId), isUrlClickable: true});
+      return true;
+    } else {
+      this.props.logger.debug(`Logo url was not replaced; entry id was not found.`);
+      this.setState({isUrlClickable: false});
+    }
+    return false;
   }
 
   /**
--- a/src/components/logo/logo.tsx
+++ b/src/components/logo/logo.tsx
@@ -64,9 +64,9 @@ class Logo extends Component<any, any> {
     const url = this.props.config.url;
     if (url && url.indexOf(ENTRY_VAR) !== -1) {
       const {player, eventManager} = this.props;
-      if (!this._maybeSetLogoUrlWithEntryId(url)) {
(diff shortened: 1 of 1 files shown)


src/components/logo/logo.tsx
```diff
@@ -64,15 +64,21 @@ class Logo extends Component<any, any> {
...
-      });
+      const entryId = player.sources.id;
+      if (typeof entryId === 'string') {
+        this.setState({urlLink: url.replace(ENTRY_VAR, entryId)});
+      } else {
+        this.setState({isUrlClickable: false});
+        eventManager.listen(player, player.Event.MEDIA_LOADED, () => {
+          const entryId = player.sources.id;
+          if (typeof entryId === 'string') {
+            this.setState({urlLink: url.replace(ENTRY_VAR, entryId), isUrlClickable: true});
```
JonathanTGold, line 75: You repeat the same logic  twice, extract all the duplicated code  into a function and reuse

JonathanTGold (changes requested):

src/components/logo/logo.tsx
```diff
@@ -64,15 +64,21 @@ class Logo extends Component<any, any> {
...
-        } else {
-          this.props.logger.debug(`Logo url was not replaced; entry id was not found.`);
-          this.setState({isUrlClickable: false});
-        }
-      });
+      const entryId = player.sources.id;
+      if (typeof entryId === 'string') {
+        this.setState({urlLink: url.replace(ENTRY_VAR, entryId)});
+      } else {
+        this.setState({isUrlClickable: false});
```
JonathanTGold, line 71: Why it is set to true from the beginning? If it depends on condition
One  you solve this , it will made the initialisation here  unnecessary

JonathanTGold (changes requested):

src/components/logo/logo.tsx
JonathanTGold, line 75: IsClickable  state is redundant (error prone) since it is always directly derived from its fellow state

JonathanTGold (changes requested):

JonathanTGold (changes requested):
Did you try the change source event ?

src/components/logo/logo.tsx
JonathanTGold: But you it is still depends on optional url config.. so  by default it should be false.. if I get it right

JonathanTGold (changes requested):

src/components/logo/logo.tsx
lianbenjamin: so you're just suggesting that it would be the opposite. starts with false and set to true in specific cases?

src/components/logo/logo.tsx
lianbenjamin: yeah, the PR was not fully ready when you started to review :) changed it already. thanks

lianbenjamin:
@JonathanTGold to your question- yes. initially I used `change_source_ended` event; the problem with change source events is that they are not being fired in case of using `setMedia` API. Hence, used `medialoaded` instead.

src/components/logo/logo.tsx
```diff
@@ -64,18 +64,33 @@ class Logo extends Component<any, any> {
...
-      eventManager.listen(player, player.Event.MEDIA_LOADED, () => {
-        const entryId = player.sources.id;
-        if (typeof entryId === 'string') {
-          this.setState({urlLink: url.replace(ENTRY_VAR, entryId)});
-        } else {
-          this.props.logger.debug(`Logo url was not replaced; entry id was not found.`);
-          this.setState({isUrlClickable: false});
-        }
-      });
+      if (!this._maybeSetLogoUrlWithEntryId(url)) {
```
SivanA-Kaltura, line 67: i think it would be clearer to have the function that sets the state and the function that tests whether the state should be set as separate functions

SivanA-Kaltura (changes requested):

SivanA-Kaltura (approved):

src/components/logo/logo.tsx
```diff
@@ -64,15 +64,30 @@ class Logo extends Component<any, any> {
...
-      eventManager.listen(player, player.Event.MEDIA_LOADED, () => {
-        const entryId = player.sources.id;
-        if (typeof entryId === 'string') {
-          this.setState({urlLink: url.replace(ENTRY_VAR, entryId)});
-        } else {
-          this.props.logger.debug(`Logo url was not replaced; entry id was not found.`);
-          this.setState({isUrlClickable: false});
-        }
-      });
+      if (!this._setLogoUrlWithEntryId(url)) {
```
JonathanTGold, line 67: Why Media_Loaded is conditionally listened
What about change media?
Change media is not limited to one during player lifetime

JonathanTGold (changes requested):

src/components/logo/logo.tsx
lianbenjamin: the logic is about adding the listener only if the entry id doesn't exist yet. there are use-cases where the entry id might exist already, before the component was mounted (like in this use-case with ima).

MosheMaorKaltura (approved):

PR #865, opened 2024-03-07
fix(SUP-41568): Thumbnail dimmed for audio track in V7

inbalvasserman:

issue:
on audio track when title is not trimmed there is black cover on all screen

fix:
change css in case audio-entry-expanded so the black cover will be only on the title

### DIFF
--- a/src/components/audio-entry-details/_audio-entry-details.scss
+++ b/src/components/audio-entry-details/_audio-entry-details.scss
@@ -55,8 +55,8 @@
 
     &.audio-entry-expanded {
       overflow: auto;
-      background: rgba(0, 0, 0, 0.7);
-      height: 100%;
+      background: linear-gradient(180deg, rgba(0, 0, 0, 0) 0%, rgba(0, 0, 0, 0.7) 100%);
+      height: auto;
 
       .audio-entry-details {
         overflow: auto;


semarche-kaltura (approved):

src/components/audio-entry-details/_audio-entry-details.scss
```diff
@@ -55,8 +55,8 @@
 
     &.audio-entry-expanded {
       overflow: auto;
-      background: rgba(0, 0, 0, 0.7);
-      height: 100%;
+      background: linear-gradient(180deg, rgba(0, 0, 0, 0) 0%, rgba(0, 0, 0, 0.7) 100%);
```
JonathanTGold, line 58: Please attach screenshot in PR desc, anyway i don’t see any requirement for gradient

JonathanTGold (changes requested):

src/components/audio-entry-details/_audio-entry-details.scss
inbalvasserman: this is the style when the size of the player is smaller. it should be like this so the part of the title will be more visible

SivanA-Kaltura (approved):

src/components/audio-entry-details/_audio-entry-details.scss
JonathanTGold, line 58: Don’t use hardcoded values , use the equivalent dynamic css var (otherwise light mode won’t be supported)

JonathanTGold (changes requested):

JonathanTGold (approved):

PR #867, opened 2024-03-20
fix(FEC-13781): add prevent seek to redux

lianbenjamin:

add new state to redux to keep `isPreventSeek` value.

### DIFF
--- a/src/reducers/seekbar.ts
+++ b/src/reducers/seekbar.ts
@@ -16,7 +16,8 @@ export const types = {
   UPDATE_SEEKBAR_SEGMENTS: `${entityName}/UPDATE_SEEKBAR_SEGMENTS`,
   UPDATE_SEGMENT_END_TIME: `${entityName}/UPDATE_SEGMENT_END_TIME`,
   ADD_SEEKBAR_CLASS: `${entityName}/ADD_SEEKBAR_CLASS`,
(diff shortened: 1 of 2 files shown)


src/components/seekbar/seekbar.tsx
```diff
@@ -28,7 +28,8 @@ const mapStateToProps = state => ({
   hidePreview: state.seekbar.hidePreview,
   hideTimeBubble: state.seekbar.hideTimeBubble,
   segments: state.seekbar.segments,
-  seekbarClasses: state.seekbar.seekbarClasses
+  seekbarClasses: state.seekbar.seekbarClasses,
+  isPreventSeek: state.seekbar.isPreventSeek
```
SivanA-Kaltura, line 32: so seekbar will be aware of preventSeek even though it's in a plugin :/
how complicated would it be to override the default seekbar by the plugin like we do in related for example ?

SivanA-Kaltura (changes requested):

src/components/seekbar/seekbar.tsx
lianbenjamin: actually, I don't want to overwrite the seekbar, i want ivq (and maybe other plugins in the future) to have access to this value and using redux seems like the best way

SivanA-Kaltura (approved):

PR #873, opened 2024-04-04
feat(FEC-13652): add meaningful error message to error overlay

lianbenjamin:

Adding meaningful information on errors.

- add new state to redux which contains `errorDetails`
  - category
  - title
  - message
- add text values of (only) the core errors in `en.i18n.json` file, per requirements
- add switch-case function which provides the title and message text keys, by the error category
- align some styles according to design


src/components/error-overlay/error-message-id-provider.ts
```diff
@@ -0,0 +1,33 @@
+type ErrorDetails = {
+  title: string;
+  message: string;
+};
+
+export const getErrorMessageIdByCategory = (errorCategory: number | undefined): ErrorDetails => {
```
SivanA-Kaltura, line 6: i think a key-value map/object of code => error details would be cleaner.
just something like errorMap.get(code) || {...//default value}

src/components/error-overlay/error-message-id-provider.ts
```diff
@@ -0,0 +1,33 @@
+type ErrorDetails = {
```
SivanA-Kaltura, line 1: i think you can rename the file to get-error-message-by-code (get-error-message-id-by-category) just so it would be clear from the filename what it contains, and put it in /utils, so the logic can be reused later

SivanA-Kaltura (changes requested):

src/components/error-overlay/error-overlay.tsx
```diff
@@ -123,8 +125,26 @@ class ErrorOverlay extends Component<any, any> {
    * @memberof ErrorOverlay
    */
   renderErrorHead(): VNode<any> | undefined {
-    const errorMessage = navigator.onLine ? 'error.default_error' : 'error.network_error';
-    return <div className={style.headline}>{this.props.errorHead ? this.props.errorHead : <Text id={errorMessage} />}</div>;
+    const {errorCategory, errorTitle, errorMessage} = this.props.errorDetails;
+    let errorTitleRes: any = '',
+      errorMessageRes: any = '';
+    if (errorTitle && errorMessage) {
```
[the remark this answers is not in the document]
lianbenjamin: hmm so I asked Product, meanwhile we assume that every errorDetails that is passed from the outside (external plugins), should provide both of them. in theory we might get only one of them but it doesn't really make sense since all errors should have title and message. however, if it does happen, the logic will get it by category id and if the category is not core, then it will fallback to "Something went wrong" generic message.

SivanA-Kaltura (approved):

PR #878, opened 2024-04-10
fix(ADA-492): Low vision users or users requiring screen to be magnified will have difficulties to read through the additional information presented

Tzipi-kaltura:

**Issue:**
When hovering on controls tooltip is displayed, when you try to move the mouse inside the tooltip area the tooltip disappear.

**Fix:**
Adding inset around the tooltip so user can move inside the tooltip before it disappear.

#### Resolves [ADA-492](https://kaltura.atlassian.net/browse/ADA-492)

[ADA-492]: https://kaltura.atlassian.net/browse/ADA-492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/tooltip/_tooltip.scss
+++ b/src/components/tooltip/_tooltip.scss
@@ -25,7 +25,7 @@
     width: max-content;
     display: table;
 
-    &:after {
+    &:before {
       content: '';
       position: absolute;
       border-width: 5px;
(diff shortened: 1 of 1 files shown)


src/components/tooltip/_tooltip.scss
```diff
@@ -117,7 +122,7 @@
       top: 50%;
       left: $tooltip-margin;
       transform: translate(0, -50%);
-      &:after {
+      &:before {
```
SivanA-Kaltura, line 125: why does replacing after with before work ?

SivanA-Kaltura (changes requested):

src/components/tooltip/_tooltip.scss
Tzipi-kaltura: because i want to add another css under the after, i added inset around the tooltip so user can move inside the tooltip before it disappear.
It does not affect the old behavior.

SivanA-Kaltura (approved):

PR #881, opened 2024-05-06
fix(FEC-13870): in and out to float causes size alignment issue

lianbenjamin:

**- bugfix**

**Issue:**
when entering to floating mode and exiting floating mode very quick (ui has not finished to update in the floating mode), the layout in the original player is sometimes off.
when the player size transitions from large to small, and before the player finishes to re-render the ui according to the new breakpoint, when transitioning back from small to large the ui still thinks it is in small size and is not being re-rendered.

**Fix:**
add a middleware for shell reducer, to manage the multiple `isSmallSize` updates, in order to prevent multiple and sometimes unnecessary updates and re-rendering.

**Related PR-** https://github.com/kaltura/playkit-js-visibility/pull/72

### DIFF
--- a/src/components/captions-menu/captions-menu.tsx
+++ b/src/components/captions-menu/captions-menu.tsx
@@ -118,7 +118,11 @@ class CaptionsMenu extends Component<any, any> {
         active: t.active,
         value: t
       }))
-      .sort((a, b) => a.label > b.label || a.label === 'Off' ? 1 : -1);
+      .map(t => ({
+        ...t,
+        label: t.label.charAt(0).toUpperCase() + t.label.slice(1)
+      }))
+      .sort((a, b) => (a.label > b.label || a.label === 'Off' ? 1 : -1));
 
     if (props.showAdvancedCaptionsMenu) {
       textOptions.push({label: props.advancedCaptionsSettingsText, value: props.advancedCaptionsSettingsText, active: false});


src/middlewares/index.ts
```diff
@@ -10,6 +11,6 @@ import {UIOptionsObject} from '../types';
  * @return {GenericStoreEnhancer} - The redux middleware.
  */
 const middleware = (player: any, config: UIOptionsObject): any =>
-  applyMiddleware<any, any>(loggerMiddleware(config), eventDispatcherMiddleware(player));
+  applyMiddleware<any, any>(loggerMiddleware(config), eventDispatcherMiddleware(player), isPlayerSmallSizeMiddleware());
```
SivanA-Kaltura, line 14: how about calling it DebounceMiddleware so it can be ready in case we want to add more actions like this ?

SivanA-Kaltura (changes requested):

src/middlewares/index.ts
lianbenjamin: done

src/middlewares/debounce-middleware.ts
```diff
@@ -0,0 +1,37 @@
+// @flow
```
JonathanTGold, line 1: can be removed..

JonathanTGold (dismissed):

SivanA-Kaltura (approved):

PR #883, opened 2024-05-12
fix(FEC-13709): remove player top bar if there is no content in it

lianbenjamin:

- bugfix.

set `visibility: hidden` to top bar component in case left and right controls do not contain any children (top bar is empty).

related PR: https://github.com/kaltura/playkit-js-ui-managers/pull/53

### DIFF
--- a/src/components/top-bar/_top-bar.scss
+++ b/src/components/top-bar/_top-bar.scss
@@ -93,6 +93,10 @@
   }
 }
 
+.player .top-bar:has(.left-controls:empty + .right-controls:empty) {
+  visibility: hidden;
+}
+
 .player.size-sm .top-bar {
   .left-controls {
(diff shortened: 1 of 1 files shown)


JonathanTGold (approved):

src/components/top-bar/_top-bar.scss
```diff
@@ -93,6 +93,10 @@
   }
 }
 
+.player .top-bar:has(.left-controls:empty + .right-controls:empty) {
+  visibility: hidden;
```
SivanA-Kaltura, line 97: Not necessarily saying this is wrong, but what about display none or changing the dimensions of the element ? Does that work ?

SivanA-Kaltura (changes requested):

src/components/top-bar/_top-bar.scss
lianbenjamin: it will probably work, but the top-bar is working with visibility property, so it made sense to keep using it.

SivanA-Kaltura (approved):

PR #884, opened 2024-05-20
fix(FEC-13880): Add css dynamic vars to support light theme

JonathanTGold:

**Issue:**  opacity element of css color of Hex format can not dynamically be changed 

**Fix:** expose the tone ramp global vars as rgb values which can be manipulated dynamically using built-in rgba() css funciton (i.e. :   `background-color: rgba(255 255 255, 0.1);)`

#### Resolves FEC-13880

Related Prs:
https://github.com/kaltura/playkit-js-share/pull/44
https://github.com/kaltura/playkit-js-downloads/pull/52

### DIFF
--- a/src/styles/dynamaic-variables.scss
+++ b/src/styles/dynamaic-variables.scss
@@ -76,6 +76,16 @@
   --playkit-tone-7-color: #212121;
   --playkit-tone-8-color: #000000;
 
+  // Tone Ramp - rgb values
+  --playkit-tone-1-color-rgb: 255, 255, 255;
+  --playkit-tone-2-color-rgb: 204, 204, 204;
+  --playkit-tone-3-color-rgb: 153, 153, 153;
+  --playkit-tone-4-color-rgb: 102, 102, 102;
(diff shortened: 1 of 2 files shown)


src/styles/dynamaic-variables.scss
```diff
@@ -76,6 +76,16 @@
   --playkit-tone-7-color: #212121;
   --playkit-tone-8-color: #000000;
 
+  // Tone Ramp - rgb values
+  --playkit-tone-1-color-rgb: 255, 255, 255;
```
SivanA-Kaltura, line 80: so what is the difference in the current color definition vs what we had before the change ?

SivanA-Kaltura (changes requested):

src/styles/dynamaic-variables.scss
JonathanTGold: before it was exported only as HEX format and now they are exported as RGB format and whenever you want more control over or be able to manipulate the color element (opacity for example ) you use this representation

read the full desc on the pr desc

SivanA-Kaltura (approved):

PR #893, opened 2024-06-19
fix(ADA-1063): Volume slider 55.00000000000001% value

JonathanTGold:

**Issue:** computer precision error in decimal numbers

**Fix:** round number


src/components/volume/volume.tsx
```diff
@@ -474,8 +474,8 @@ class Volume extends Component<any, any> {
           role="slider"
           aria-valuemin={0}
           aria-valuemax={100}
-          aria-valuenow={player.volume * 100}
-          aria-valuetext={`${player.volume * 100}% volume ${player.muted ? 'muted' : ''}`}>
+          aria-valuenow={Math.round(player.volume * 100)}
```
semarche-kaltura, line 477: I would create `volume` const 
```
const volume = Math.round(player.volume * 100)
```
and then use it in `aria-valuenow` and `aria-valuetext` instead of execute Math.round 2 times

SivanA-Kaltura (approved):

PR #896, opened 2024-06-26
ADA-1065-Added aria-labelby attribute to the dialog

RazvanBrinzoiK05:

**Issue:**

Accessibility change that was required for announcing the dialog according to it's title.

**Fix:**

Added aria-labelledby to associate the dialog with the descriptive title, improving accessibility for screen reader users.

#### Resolves [ADA-1065](https://kaltura.atlassian.net/browse/ADA-1065)

[ADA-1065]: https://kaltura.atlassian.net/browse/ADA-1065?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/cvaa-overlay/cvaa-overlay.tsx
+++ b/src/components/cvaa-overlay/cvaa-overlay.tsx
@@ -131,6 +131,9 @@ class CVAAOverlay extends Component<any, any> {
    */
   render(props: any): VNode<any> {
     props.clearAccessibleChildren();
+    const isMainOverlay = this.state.activeWindow === cvaaOverlayState.Main;
+    const ariaProps = isMainOverlay ? { ariaLabelledBy: 'captionsDialogTitle' } : { ariaLabel: 'Custom Captions Settings' };
+
     return (
       <Overlay
         open
@@ -138,6 +141,7 @@ class CVAAOverlay extends Component<any, any> {
         addAccessibleChild={this.props.addAccessibleChild}
         onClose={props.onClose}
         type="cvaa"
+        {...ariaProps}
       >
         {this.state.activeWindow === cvaaOverlayState.Main ? (
           <MainCaptionsWindow
--- a/src/components/cvaa-overlay/cvaa-overlay.tsx
+++ b/src/components/cvaa-overlay/cvaa-overlay.tsx
@@ -11,6 +11,7 @@ import {withLogger} from '../logger';
 import {MainCaptionsWindow} from './main-captions_window';
 import {CustomCaptionsWindow} from './custom-captions-window';
 import {withText} from 'preact-i18n';
+import { Utils } from '@playkit-js/playkit-js';
 
 /**
  * mapping state to props
@@ -41,7 +42,10 @@ const COMPONENT_NAME = 'CVAAOverlay';
 @withLogger(COMPONENT_NAME)
 @withEventDispatcher(COMPONENT_NAME)
 @withKeyboardA11y
-@withText({cvvaDialogText: 'cvaa.title'})
+@withText({
+  cvvaDialogText: 'cvaa.title',
+  cvvaSetCustomCaptionsText: 'cvaa.set_custom_caption',
+})
 class CVAAOverlay extends Component<any, any> {
   /**
    * componentWillUnmount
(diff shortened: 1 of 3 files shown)


src/components/cvaa-overlay/main-captions_window.tsx
```diff
@@ -75,7 +75,7 @@ class MainCaptionsWindow extends Component<any, any> {
   render(props: any): VNode<any> {
     return (
       <div className={[style.overlayScreen, style.active].join(' ')}>
-        <h2 className={style.title}>
+        <h2 className={style.title} id="captionsDialogTitle">
```
SivanA-Kaltura, line 78: is there a way to avoid using id ?
id should be unique for the page, but we can have multiple players on a single page

src/components/overlay/overlay.tsx
```diff
@@ -140,7 +140,7 @@ class Overlay extends Component<any, any> {
     }
 
     return (
-      <div tabIndex={-1} className={overlayClass.join(' ')} role="dialog" onKeyDown={this.onKeyDown} aria-label={label}>
+      <div tabIndex={-1} className={overlayClass.join(' ')} role="dialog" onKeyDown={this.onKeyDown} aria-labelledby="captionsDialogTitle">
```
SivanA-Kaltura, line 143: does tooltip still work if aria-label is removed ?

src/components/cvaa-overlay/cvaa-overlay.tsx
```diff
@@ -138,7 +138,6 @@ class CVAAOverlay extends Component<any, any> {
         addAccessibleChild={this.props.addAccessibleChild}
         onClose={props.onClose}
         type="cvaa"
-        label={props.cvvaDialogText}
```
SivanA-Kaltura, line 141: why remove it ?

SivanA-Kaltura (changes requested):

src/components/cvaa-overlay/main-captions_window.tsx
RazvanBrinzoiK05: Unfortunately the dialog has a visible title and it must be referenced with aria-labelledby that must point to the title id. If the title would be missing we could've used just an aria-label property that doesn't use id.

src/components/overlay/overlay.tsx
RazvanBrinzoiK05: Yes, works the same as before.

src/components/cvaa-overlay/cvaa-overlay.tsx
RazvanBrinzoiK05: The label was passed as a prop just for the aria-label property that was replaced.

RazvanBrinzoiK05:
Considered adding an aria property when custom caption settings is rendered as well.

src/components/overlay/overlay.tsx
```diff
@@ -140,7 +140,7 @@ class Overlay extends Component<any, any> {
     }
 
     return (
-      <div tabIndex={-1} className={overlayClass.join(' ')} role="dialog" onKeyDown={this.onKeyDown} aria-label={label}>
+      <div tabIndex={-1} className={overlayClass.join(' ')} role="dialog" onKeyDown={this.onKeyDown} {...(ariaLabelledBy ? { 'aria-labelledby': ariaLabelledBy } : { 'aria-label': ariaLabel })}>
```
SivanA-Kaltura, line 143: i think putting the ternary operator outside the return statement, like you did in cvaa-overlay, is more readable

SivanA-Kaltura (changes requested):

src/components/cvaa-overlay/cvaa-overlay.tsx
```diff
@@ -131,14 +131,17 @@ class CVAAOverlay extends Component<any, any> {
    */
   render(props: any): VNode<any> {
     props.clearAccessibleChildren();
+    const isMainOverlay = this.state.activeWindow === cvaaOverlayState.Main;
+    const ariaProps = isMainOverlay ? { ariaLabelledBy: 'captionsDialogTitle' } : { ariaLabel: 'Custom Captions Settings' };
```
SivanA-Kaltura, line 135: We usually don't use hardcoded english texts.
Is this a text that doesn't need translation ? And non of the existing cvaa localized texts can be used here.

If this text is new, and doesn't exist anywhere yet in the ui (so it's not possible to reuse the localized version), it should be added to an en.18n.json file under "cvaa" section, and product should be notified that this text requires localization, so it can be translated in the next opportunity.

For examples of localization you can look at uses of withText decorator in playkit-ui.

SivanA-Kaltura (changes requested):

SivanA-Kaltura (approved):

PR #898, opened 2024-06-27
feat(FEC-13945): move core controls to upper bar

lianbenjamin:

**New requirement:**
if there is not enough space for controls in bottom bar, instead of removing them- move them to the upper bar.
This was requested for A11y purposes.

- keep 10sec FWD and BWD controls in bottom bar (do not remove them)
- new service to register relevant core components
- new redux state to update the controls that need to be moved from bottom bar to upper bar
- new interface that registers the components to the new service
- a special implementation for `TimeDisplay` component (different behavior)

related PR: https://github.com/kaltura/playkit-js-ui-managers/pull/59

### DIFF
--- a/src/components/advanced-audio-desc/advanced-audio-desc.tsx
+++ b/src/components/advanced-audio-desc/advanced-audio-desc.tsx
@@ -9,6 +9,9 @@ import {connect} from 'react-redux';
 import {ButtonControl} from '../button-control';
 import {bindActions} from '../../utils';
 import {actions} from '../../reducers/settings';
+import {IconComponent, registerToBottomBar} from '../bottom-bar';
+import {redux} from '../../index';
+import {withPlayer} from '../player';
 
 const COMPONENT_NAME = 'AdvancedAudioDesc';
 
@@ -29,12 +32,38 @@ const mapStateToProps = state => ({
  * @extends {Component}
  */
 @connect(mapStateToProps, bindActions(actions))
+@withPlayer
 @withEventDispatcher(COMPONENT_NAME)
 @withText({AdvancedAudioDescText: 'settings.advancedAudioDescription'})
-class AdvancedAudioDesc extends Component<any, any> {
+class AdvancedAudioDesc extends Component<any, any> implements IconComponent {
   constructor(props: any) {
     super();
     this.state = {toggle: false};
+    registerToBottomBar(COMPONENT_NAME, props.player, () => this.registerComponent());
+  }
+
+  registerComponent(): any {
+    return {
+      ariaLabel: () => this.getComponentText(),
+      displayName: COMPONENT_NAME,
+      order: 5,
+      svgIcon: () => this.getSvgIcon(),
+      onClick: () => this.onClick(),
+      component: () => {
+        return getComp({...this.props, classNames: [style.upperBarIcon]});
+      },
+      selfManagement: true
+    };
+  }
+
+  getComponentText = (): any => {
+    return this.props.AdvancedAudioDescText;
+  }
+
+  getSvgIcon = (): any => {
+    return {
+      type: redux.useStore().getState().settings.advancedAudioDesc ? IconType.AdvancedAudioDescriptionActive : IconType.AdvancedAudioDescription
+    };
   }
 
   /**
@@ -69,7 +98,7 @@ class AdvancedAudioDesc extends Component<any, any> {
    */
   render({AdvancedAudioDescText, innerRef}: any): VNode<any> | undefined {
     return !this._shouldRender() ? undefined : (
-      <ButtonControl name={COMPONENT_NAME} className={style.noIdleControl}>
+      <ButtonControl name={COMPONENT_NAME} className={[style.noIdleControl, this.props.classNames ? this.props.classNames.join(' ') : ''].join(' ')}>
         <Tooltip label={AdvancedAudioDescText}>
           <Button tabIndex="0" aria-label={AdvancedAudioDescText} className={`${style.controlButton}`} ref={innerRef} onClick={this.onClick}>
             <Icon type={this.state.toggle ? IconType.AdvancedAudioDescriptionActive : IconType.AdvancedAudioDescription} />
@@ -78,6 +107,10 @@ class AdvancedAudioDesc extends Component<any, any> {
       </ButtonControl>
     );
   }
+}
+
+const getComp = (props: any): VNode => {
+  return <AdvancedAudioDesc {...props} />;
 }
 
 AdvancedAudioDesc.displayName = COMPONENT_NAME;
--- a/src/components/advanced-audio-desc/advanced-audio-desc.tsx
+++ b/src/components/advanced-audio-desc/advanced-audio-desc.tsx
@@ -50,9 +50,9 @@ class AdvancedAudioDesc extends Component<any, any> implements IconComponent {
       svgIcon: () => this.getSvgIcon(),
       onClick: () => this.onClick(),
       component: () => {
-        return getComp({...this.props, classNames: [style.upperBarIcon]});
+        return getComponent({...this.props, classNames: [style.upperBarIcon]});
       },
-      selfManagement: true
+      shouldHandleOnClick: false
     };
   }
 
@@ -109,7 +109,7 @@ class AdvancedAudioDesc extends Component<any, any> implements IconComponent {
   }
 }
 
-const getComp = (props: any): VNode => {
+const getComponent = (props: any): VNode => {
   return <AdvancedAudioDesc {...props} />;
 }
 
--- a/src/components/bottom-bar/_bottom-bar.scss
+++ b/src/components/bottom-bar/_bottom-bar.scss
@@ -49,6 +49,15 @@
   .bottom-bar-area {
     position: relative;
     pointer-events: none;
+
+    .time-display {
+      padding: 0;
+      position: absolute;
+      bottom: 20px;
+      font-size: 12px;
+      line-height: normal;
+      font-weight: 400;
+    }
   }
 
   .control-button-container {
@@ -66,16 +75,10 @@
       margin: 0;
     }
   }
-  &.size-xs {
-    .left-controls {
-      .control-button-container {
-        display: none;
-      }
-    }
-  }
+
   &.size-xs,
   &.size-sm {
-    .control-button-container {
+    .control-button-container:not(.upper-bar-icon) {
       margin: 0 3px;
     }
   }
--- a/src/components/bottom-bar/bottom-bar-registry-manager.ts
+++ b/src/components/bottom-bar/bottom-bar-registry-manager.ts
@@ -12,7 +12,7 @@ export class BottomBarRegistry {
   }
 
   public register(component: string, componentIcon: any): void {
-    if(!this._registry.get(component)) {
+    if (!this._registry.get(component)) {
       this._registry.set(component, componentIcon);
     }
   }
@@ -35,8 +35,8 @@ export class BottomBarRegistry {
 export const bottomBarRegistryManager: string = 'bottomBarRegistryManager';
 
 export const registerToBottomBar = (compName: string, player: any, getIconDtoCallback: () => any): void => {
-  const bottomBarRegistry = player?.getService(bottomBarRegistryManager) as BottomBarRegistry || undefined;
+  const bottomBarRegistry = (player?.getService(bottomBarRegistryManager) as BottomBarRegistry) || undefined;
   if (bottomBarRegistry && !bottomBarRegistry.registry.get(compName)) {
     bottomBarRegistry.register(compName, getIconDtoCallback());
   }
-}
+};
--- a/src/components/bottom-bar/bottom-bar-registry-manager.ts
+++ b/src/components/bottom-bar/bottom-bar-registry-manager.ts
@@ -1,0 +1,42 @@
+export interface IconComponent {
+  registerComponent(): any;
+  getSvgIcon(): any;
+  getComponentText(): any;
+}
+
+export class BottomBarRegistry {
+  private _registry: Map<string, any>;
+
+  constructor() {
+    this._registry = new Map();
+  }
+
+  public register(component: string, componentIcon: any): void {
+    if(!this._registry.get(component)) {
+      this._registry.set(component, componentIcon);
+    }
+  }
+
+  public unregister(component: string): void {
+    if (this._registry.get(component)) {
+      this._registry.delete(component);
+    }
+  }
+
+  public get registry(): Map<string, any> {
+    return this._registry;
+  }
+
+  public clear(): void {
+    this._registry.clear();
+  }
+}
+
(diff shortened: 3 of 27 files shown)


src/components/picture-in-picture/picture-in-picture.tsx
```diff
@@ -56,6 +57,39 @@ class PictureInPicture extends Component<any, any> {
...
+
+  registerComponent(): any {
+    return {
+      ariaLabel: () => this.getComponentText(),
+      displayName: COMPONENT_NAME,
+      order: 5,
+      svgIcon: () => this.getSvgIcon(),
+      onClick: () => this.togglePip(),
+      component: () => {
+        return getComp({...this.props, classNames: [style.upperBarIcon]});
```
SivanA-Kaltura, line 77: i dont like how all the classes have the same getComp function, but on the other hand if i remember right we can't put it on an interface.

does the syntax 

return PictureInPicture({...this.props, classNames: [style.upperBarIcon]}) work here ?
if not, i'd prefer to somehow have reuse instead of again, having very similar functions
it's also possible to have an entirely separate class which will only have a function which returns <PictureInPicture>, and then have one such class per every component, but while more organized it seems a bit overkill to me to have it for one single method...

src/components/engine-connector/engine-connector.tsx
```diff
@@ -193,6 +195,7 @@ class EngineConnector extends Component<EngineConnectorProps, any> {
     eventManager.listen(player, player.Event.Core.TEXT_TRACK_CHANGED, () => {
       let tracks = player.getTracks(TrackType.TEXT);
       this.props.updateTextTracks(tracks);
+      this.props.updateIsTextOn(tracks.find(track => track.active)?.language !== 'off');
```
SivanA-Kaltura, line 198: as mentioned below
i think subtitles_enabled / disabled would be clearer

src/components/closed-captions/closed-captions.tsx
```diff
@@ -43,8 +48,50 @@ const ClosedCaptions = connect(mapStateToProps)(
...
+        return {
+          ariaLabel: () => getAriaLabel(),
+          displayName: COMPONENT_NAME,
+          order: 5,
+          svgIcon: () => getSvgIcon(),
+          onClick: () => onClick(),
+          component: () => {
+            return getComp({...props, classNames: [style.upperBarIcon]});
+          },
+          selfManagement: true
```
SivanA-Kaltura, line 65: the name "selfManagement" is not very clear (actually "selfManaged" is already clearer). what is the value used for ?

src/components/captions-control/captions-control-mini.tsx
```diff
@@ -0,0 +1,112 @@
...
+      },
+      selfManagement: true
+    };
+  };
+
+  useEffect(() => {
+    setShouldRender(controlsToMove.includes(CaptionsControl.displayName));
+  }, [controlsToMove]);
+
+  const getIsTextOnFromStore = (): boolean => {
```
SivanA-Kaltura, line 53: i feel like in a functional component there's probably a nicer way to wrap up the getter of this value
maybe with useMemo

src/components/captions-control/captions-control-mini.tsx
```diff
@@ -0,0 +1,112 @@
...
+  controlsToMove: state.bottomBar.controlsToMove
+});
+
+const COMPONENT_NAME = 'CaptionsControlMini';
+const CaptionsControlMini = connect(mapStateToProps)((props, context) => {
+  const [shouldRender, setShouldRender] = useState(false);
+
+  const {player} = context;
+  const {controlsToMove} = props;
+  const targetId: HTMLDivElement | Document = (document.getElementById(player.config.targetId) as HTMLDivElement) || document;
```
SivanA-Kaltura, line 27: think of a better name, since the variable holds a div and not an id.
we should actually have a utility function for this, i don't think components should try to access the div / dom by themselves.
not crucial here though. just for the name to be correct.

src/components/captions-control/captions-control-mini.tsx
```diff
@@ -0,0 +1,112 @@
+import {createPortal, useEffect} from 'preact/compat';
```
SivanA-Kaltura, line 1: huh. where did the "/compat" come from ?

src/components/advanced-audio-desc/advanced-audio-desc.tsx
```diff
@@ -29,12 +32,38 @@ const mapStateToProps = state => ({
...
     super();
     this.state = {toggle: false};
+    registerToBottomBar(COMPONENT_NAME, props.player, () => this.registerComponent());
+  }
+
+  registerComponent(): any {
+    return {
+      ariaLabel: () => this.getComponentText(),
+      displayName: COMPONENT_NAME,
+      order: 5,
```
SivanA-Kaltura, line 49: didnt we want to put the order in some global place ?

src/components/advanced-audio-desc/advanced-audio-desc.tsx
```diff
@@ -29,12 +32,38 @@ const mapStateToProps = state => ({
...
+    };
+  }
+
+  getComponentText = (): any => {
+    return this.props.AdvancedAudioDescText;
+  }
+
+  getSvgIcon = (): any => {
+    return {
+      type: redux.useStore().getState().settings.advancedAudioDesc ? IconType.AdvancedAudioDescriptionActive : IconType.AdvancedAudioDescription
```
SivanA-Kaltura, line 65: there's probably a better way to fetch value from state...
check what the best practice is (for example, theres an option to have a file full of getters that just return stuff from state). also, maybe getMemo can be used here.

src/components/bottom-bar/bottom-bar-registry-manager.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  public unregister(component: string): void {
+    if (this._registry.get(component)) {
+      this._registry.delete(component);
+    }
+  }
+
+  public get registry(): Map<string, any> {
+    return this._registry;
```
SivanA-Kaltura, line 27: if register and unregister are public, why is registry also public ?

SivanA-Kaltura (changes requested):

src/components/tooltip/_tooltip.scss
```diff
@@ -137,6 +137,7 @@
   &.show {
     visibility: visible;
     opacity: 1;
+    z-index: 1;
```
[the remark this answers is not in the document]
lianbenjamin: hmm actually i don't think it can break anything, because either way tooltip should be on top of everything if i think about it

src/components/scrollable/scrollable.tsx
```diff
@@ -36,11 +37,24 @@ const Scrollable = ({children, isVertical}: {children: ComponentChildren; isVert
     }
   };
 
+  useLayoutEffect(() => {
+    if (isVertical) {
+      const scrollableEl = document.getElementsByClassName('playkit-scrollable')[0];
```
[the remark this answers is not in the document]
lianbenjamin: yep, will do

src/components/picture-in-picture/picture-in-picture.tsx
lianbenjamin: update- as discussed, leaving it the way it is now (just refactoring the name to `getComponent`)

src/components/closed-captions/closed-captions.tsx
lianbenjamin: renamed to `shouldHandleOnClick`

src/components/captions-control/captions-control-mini.tsx
lianbenjamin: hmm because of the `createPortal`

src/components/bottom-bar/bottom-bar-registry-manager.ts
lianbenjamin: register just adds the component to the map. in ui-managers when we need to add a component to the upper-bar, we need to get the component from the registry itself (the map).

src/components/captions-control/captions-control-mini.tsx
SivanA-Kaltura: I mean, we usually import this from "preact"
How come this came out different ?
"compat" has createPortal and normal preact doesn't ? why ?

SivanA-Kaltura (changes requested):

src/components/bottom-bar/bottom-bar-registry-manager.ts
SivanA-Kaltura: isn't that what get() does ? return a component from the map ?

src/components/tooltip/_tooltip.scss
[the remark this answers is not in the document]
SivanA-Kaltura: please check with sergey because he made some changes in the tooltip z index area

src/components/captions-control/captions-control-mini.tsx
lianbenjamin: I'm not sure, I took the same import we do in closed captions component..

src/components/bottom-bar/bottom-bar-registry-manager.ts
lianbenjamin: but you need to apply the `get()` on one of the items... to get the item you need the `registry`

src/components/captions-control/captions-control-mini.tsx
lianbenjamin: @SivanA-Kaltura wdym "normal preact"?
so we have `preact/hooks` which exports the hooks (`useEffect`, `useMemo`, etc...) and then there's `preact/compat` which exports the `createPortal` and apparently also the hooks from `preact/hooks`, see the code line below from `preact/compat`:
`export import useEffect = _hooks.useEffect;`

To sum up:
- `createPortal` can be imported **only** from `preact/compat`
- while the hooks can be imported from `preact/compat` **and** `preact/hooks`

SivanA-Kaltura (approved):

src/components/captions-control/captions-control-mini.tsx
```diff
@@ -0,0 +1,112 @@
...
+import {Text} from 'preact-i18n';
+import {CaptionsMenu} from '../captions-menu';
+import {SmartContainer} from '../smart-container';
+import {CaptionsControl} from './captions-control';
+
+const mapStateToProps = state => ({
+  controlsToMove: state.bottomBar.controlsToMove
+});
+
+const COMPONENT_NAME = 'CaptionsControlMini';
```
JonathanTGold, line 21: can you elaborate why do we need to duplicate the caption component

JonathanTGold (changes requested):

src/components/vr-stereo/vr-stereo.tsx
```diff
@@ -38,7 +40,40 @@ const COMPONENT_NAME = 'VrStereo';
...
+    return {
+      ariaLabel: () => this.getComponentText(),
+      displayName: COMPONENT_NAME,
+      order: 5,
+      svgIcon: () => this.getSvgIcon(),
+      onClick: () => this.onClick(),
+      component: () => {
+        return getComponent({...this.props, classNames: [style.upperBarIcon]});
+      },
+      shouldHandleOnClick: false
```
JonathanTGold, line 63: why the onClick is not sufficient to deduce is it should be handheld

JonathanTGold (changes requested):

src/components/bottom-bar/_bottom-bar.scss
```diff
@@ -49,6 +49,15 @@
   .bottom-bar-area {
     position: relative;
     pointer-events: none;
+
+    .time-display {
```
JonathanTGold, line 53: I think time-display CSS should not be in bottom bar components

JonathanTGold (changes requested):

src/components/tooltip/tooltip.tsx
```diff
@@ -17,6 +17,7 @@ interface TooltipOwnProps {
   maxWidth?: string;
   label: string;
   strictPosition?: boolean;
+  className?: string;
```
JonathanTGold, line 20: this is too generic name, its role is not clear

JonathanTGold (changes requested):

src/components/bottom-bar/bottom-bar-registry-manager.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  public clear(): void {
+    this._registry.clear();
+  }
+}
+
+export const bottomBarRegistryManager: string = 'bottomBarRegistryManager';
+
+export const registerToBottomBar = (componentName: string, player: any, getIconDtoCallback: () => any): void => {
```
JonathanTGold, line 37: why the registration need to be exported as a separate function and not as a singleton service as we do in upperBarManager

JonathanTGold (changes requested):

src/components/captions-control/captions-control-mini.tsx
```diff
@@ -0,0 +1,112 @@
...
+      get: () => {
+        return (
+        <SmartContainer targetId={player.config.targetId} onClose={() => removeCaptionsOverlay()} title={<Text id="controls.language" />}>
+          <CaptionsMenu asDropdown={false} onAdvancedCaptionsClick={openCVAAOverlay}/>
+        </SmartContainer>
+        )}
+    });
+  };
+
+  return shouldRender ? (
```
JonathanTGold, line 91: insted of generic should render it is better to name the reason that prevent rendering

JonathanTGold (changes requested):

src/components/vr-stereo/vr-stereo.tsx
lianbenjamin: as explained here- https://github.com/kaltura/playkit-js-ui-managers/pull/59#discussion_r1673482830

src/components/tooltip/tooltip.tsx
lianbenjamin: the role is className, we do the same in other places where we want to pass calssNames as props.

src/components/bottom-bar/bottom-bar-registry-manager.ts
lianbenjamin: can't remember why Singleton wasn't a good option. i think in case we have multiple players on page..

src/components/bottom-bar/_bottom-bar.scss
lianbenjamin: in some cases, time-display component is becoming part of the bottom-bar area.

src/components/captions-control/captions-control-mini.tsx
lianbenjamin: since the `captionsControl` itself did not work as expected in the upper bar (while all the others are working just fine), the solution that was agreed is to add another "mini" component of it that would be rendered in the upper bar instead of the `captionsControl`, whenever it's needed.

MosheMaorKaltura (approved):

PR #907, opened 2024-07-19
fix(ADA-1275): When user focuses the player he will be able to play/pause with "space" key

RazvanBrinzoiK05:
**Issue:**
After the pre-playback-overlay is rendered, with the centered big play button shown in the player container, the user can use it to start the playback and make the overlay component disappear. Afterwards he can play/pause the media by pressing "Space" key due to the fact that initially the play/pause button is focused.

The issue appears when the user switches focus from the player and returns to it, because the play button will not be focused anymore.

**Fix:**

Added event listener on the player container for the space key.

### DIFF
--- a/src/components/play-pause/play-pause.tsx
+++ b/src/components/play-pause/play-pause.tsx
@@ -63,6 +63,8 @@ class PlayPause extends Component<any, any> {
         this._playPauseButtonRef?.focus();
       });
     });
+    const myPlayerContainer = document.querySelector('.kaltura-player-container');
+    eventManager.listen(myPlayerContainer,'keydown', this.handleSpaceKey);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/components/play-pause/play-pause.tsx
```diff
@@ -63,6 +63,8 @@ class PlayPause extends Component<any, any> {
         this._playPauseButtonRef?.focus();
       });
     });
+    const myPlayerContainer = document.querySelector('.kaltura-player-container');
```
semarche-kaltura, line 66: there might be more than one instance of Kaltura-player on page. You have to specify exact ID of player in selector. Id of player can be get from player configuration (targetId)

semarche-kaltura (changes requested):

semarche-kaltura (approved):

PR #909, opened 2024-07-24
fix(FEC-14059): fix the height of scrollable component

lianbenjamin:

bugfix - regression.

**Issue:**
recent changes in `scrollable` component have caused `More` dropdown to not change its height dynamically.
the root cause is the new `height` style that was added, which exists in cases where it shouldn't.

**Fix:**
- make an accurate calculation to check if there is enough space for the scrollable items inside the parent's container
- set `height` and remove `height` style according to the calculation

### DIFF
--- a/src/components/scrollable/scrollable.scss
+++ b/src/components/scrollable/scrollable.scss
@@ -2,14 +2,26 @@
   display: flex;
   width: 100%;
 
+  .items-container {
+    display: flex;
+    width: 100%;
+    height: fit-content;
+  }
+
   &.horizontal {
     flex-direction: row;
     overflow: auto hidden;
+    .items-container {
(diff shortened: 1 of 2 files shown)


semarche-kaltura (approved):

src/components/scrollable/scrollable.scss
```diff
@@ -2,14 +2,26 @@
   display: flex;
   width: 100%;
 
+  .items-container {
+    display: flex;
+    width: 100%;
+    height: fit-content;
```
SivanA-Kaltura, line 8: does it work with other values ?

SivanA-Kaltura (changes requested):

SivanA-Kaltura (approved):

PR #915, opened 2024-07-30
ADA-1499 - Make aria label equal to logo title

valentinStoicaK:

For the Logo link the SR was not announcing a meaningful description. The code was modified so that the message announced can now be customized by the client by modifying the Logo Title.

**Issue:**
Logo link aria-label is not meaningful
**Fix:**
Logo link aria-label is equal to the logo title or absence with a default title

### DIFF
--- a/src/components/logo/logo.tsx
+++ b/src/components/logo/logo.tsx
@@ -126,7 +126,7 @@ class Logo extends Component<any, any> {
           onClick={this._handleOnClick}
           className={style.controlButton}
           href={this.state.urlLink}
-          aria-label={props.logoText}
+          aria-label={props.config.text ? props.config.text : "Logo"}
           target="_blank"
           rel="noopener noreferrer"
           tabIndex={0}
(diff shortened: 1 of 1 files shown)


src/components/logo/logo.tsx
```diff
@@ -126,7 +126,7 @@ class Logo extends Component<any, any> {
           onClick={this._handleOnClick}
           className={style.controlButton}
           href={this.state.urlLink}
-          aria-label={props.logoText}
+          aria-label={props.config.text ? props.config.text : "Logo"}
```
SivanA-Kaltura, line 129: first of all, please update the PR description.

second, props.logoText is localized, while the hardcoded "Logo" is not. did you mean to do something like aria-label={props.config.text || props.logoText} ?

SivanA-Kaltura (changes requested):

src/components/logo/logo.tsx
valentinStoicaK: Yes, you are right, it's better this way

valentinStoicaK: Done

SivanA-Kaltura (approved):

PR #916, opened 2024-08-01
fix(FEC-14076): Player v7 | Download | Download notification popup is being display behind download overlay.

SivanA-Kaltura:

Move interactive area filter to a separate element
If interactive area itself has a filter, it affects every component within the interactive area, including popups, which should not be affected by it

Related PR: https://github.com/kaltura/playkit-js-ui-managers/pull/61

Resolves FEC-14076

**Issue:**

**Fix:**

### DIFF
--- a/src/components/interactive-area/_interactive-area.scss
+++ b/src/components/interactive-area/_interactive-area.scss
@@ -1,5 +1,9 @@
-.player.overlay-active .interactive-area {
-  filter: blur($blur);
+.player.overlay-active .interactive-area {  
+  .filter {
+    filter: blur($blur);
+    height: 100%;
+    width: 100%;
+  }
 }
 .interactive-area {
   pointer-events: none;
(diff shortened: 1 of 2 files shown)


lianbenjamin (approved):

src/components/interactive-area/_interactive-area.scss
```diff
@@ -1,5 +1,9 @@
-.player.overlay-active .interactive-area {
-  filter: blur($blur);
+.player.overlay-active .interactive-area {  
+  .filter {
+    filter: blur($blur);
```
JonathanTGold, line 3: this is the core component it should be agnostic and Un-Opinionated

JonathanTGold (changes requested):

src/components/interactive-area/interactive-area.tsx
```diff
@@ -26,6 +26,7 @@ class InteractiveArea extends Component<any, any> {
     const {children} = this.props;
     return (
       <div className={style.interactiveArea}>
+        <div className={style.filter}></div>
```
JonathanTGold, line 29: core should not apply style

JonathanTGold (changes requested):

PR #917, opened 2024-08-02
fix(ADA-1404): Focus handling on pre-playback

RazvanBrinzoiK05:
Issue:
On a first render in case a user is navigating with Tab key, the element that receives the first focus is the pre-playback-play-button that has a default styling of "outline: none ". 

Fix:
Adding existing outline styling for this particular case of focus as well.

### DIFF
--- a/src/styles/_buttons.scss
+++ b/src/styles/_buttons.scss
@@ -44,6 +44,10 @@
   padding: 2px;
 }
 
+.pre-playback-play-overlay button.pre-playback-play-button:focus{
+  outline: 1px solid $tab-focus-color;
+}
+
 .player {
   button {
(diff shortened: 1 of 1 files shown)


src/styles/_buttons.scss
```diff
@@ -44,6 +44,10 @@
   padding: 2px;
 }
 
+.pre-playback-play-overlay button.pre-playback-play-button:focus{
+  outline: 1px solid $tab-focus-color;
```
SivanA-Kaltura, line 48: hi, did you check with Dana about this style ?

SivanA-Kaltura (changes requested):

PR #922, opened 2024-08-14
ADA-1572 - Replace a tag with span when link is missing

valentinStoicaK:
When the link is not provided the anchor tag is replaced with a span tag

**Issue:**
The SR was announcing logo as link even when no link was provided
**Fix:**

### DIFF
--- a/src/components/logo/logo.tsx
+++ b/src/components/logo/logo.tsx
@@ -138,7 +138,6 @@ class Logo extends Component<any, any> {
           <span
             className={style.controlButton}
             aria-label={props.config.text || props.logoText}
-            tabIndex={0}
           >
             <img className={style.icon} src={props.config.img} />
           </span>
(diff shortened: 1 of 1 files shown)


src/components/logo/logo.tsx
```diff
@@ -133,9 +134,19 @@ class Logo extends Component<any, any> {
           >
           <img className={style.icon} src={props.config.img} />
         </a>
+        ) : (
+          <span
+            className={style.controlButton}
+            aria-label={props.config.text || props.logoText}
+            tabIndex={0}
```
SivanA-Kaltura, line 141: why is there a tab index if the element is not clickable ?

SivanA-Kaltura (changes requested):

src/components/logo/logo.tsx
valentinStoicaK: You are right, the element doesn't need to be navigable any more

SivanA-Kaltura (approved):

PR #926, opened 2024-08-19
fix(FEC-14108): handle keydown on logo component

semarche-kaltura:

resolves: https://kaltura.atlassian.net/browse/FEC-14108

### DIFF
--- a/src/components/logo/logo.tsx
+++ b/src/components/logo/logo.tsx
@@ -6,6 +6,7 @@ import {withPlayer} from '../player';
 import {withLogger} from '../logger';
 import {withEventManager} from '../../event';
 import {withEventDispatcher} from '../event-dispatcher';
+import {KeyMap} from '../../utils';
 
 const COMPONENT_NAME = 'Logo';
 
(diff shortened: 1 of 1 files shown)


src/components/logo/logo.tsx
```diff
@@ -74,6 +75,14 @@ class Logo extends Component<any, any> {
     }
   }
 
+  _handleKeyDown = (e: KeyboardEvent): void => {
+    if (e.keyCode === KeyMap.ENTER || e.keyCode === KeyMap.SPACE) {
+      window.open(this.state.urlLink, '_blank');
```
SivanA-Kaltura, line 80: what if we are in an iframe ?

SivanA-Kaltura (changes requested):

src/components/logo/logo.tsx
semarche-kaltura: the solution works fine, example from ui-managers
https://github.com/kaltura/playkit-js-ui-managers/blob/3b2567a5093ae94b11b7bc38c5ab08b4fcf2bd37/src/services/side-panels-manager/models/item-wrapper.tsx#L81

PR #932, opened 2024-08-27
fix(ADA-1610): Update aria label for volume button

valentinStoicaK:

- Removed slider aria-label from volume button;
- Added aria-description to slider so the user would be informed about slider controls;

**Issue:**

**Fix:**

### DIFF
--- a/src/components/volume/volume.tsx
+++ b/src/components/volume/volume.tsx
@@ -49,7 +49,7 @@ const KEYBOARD_DEFAULT_VOLUME_JUMP: number = 5;
 const translates = (props: any) => ({
   volumeLabel: props.muted ? <Text id="controls.unmute">Unmute</Text> : <Text id="controls.mute">Mute</Text>,
   sliderAriaLabel: <Text id="volume.volume_slider_aria_label">Volume</Text>, 
-  sliderAriaDescription: <Text id="volume.volume_slider_aria_label">Use the arrows to control the volume</Text>
+  sliderAriaDescription: <Text id="volume.volume_slider_description">Use the arrows to control the volume</Text>
 });
 
 /**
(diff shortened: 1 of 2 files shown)


src/components/volume/volume.tsx
```diff
@@ -48,7 +48,8 @@ const KEYBOARD_DEFAULT_VOLUME_JUMP: number = 5;
  */
 const translates = (props: any) => ({
   volumeLabel: props.muted ? <Text id="controls.unmute">Unmute</Text> : <Text id="controls.mute">Mute</Text>,
-  sliderAriaLabel: <Text id="volume.volume_slider_aria_label">Volume control, use the arrows to control the volume</Text>
+  sliderAriaLabel: <Text id="volume.volume_slider_aria_label">Volume</Text>, 
+  sliderAriaDescription: <Text id="volume.volume_slider_aria_label">Use the arrows to control the volume</Text>
```
SivanA-Kaltura, line 52: The value inside text is the default english text
The actual localized value is coming from the key-value pair in id field

Please check with product whether they want to take the existing translations we have and split them into two separate fields within the i18n files, or whether we need a new translation for the separate values

Meaning, in french for example, we can split the values from

"volume_slider_aria_label": "Contrôle du volume, utiliser les flèches pour contrôler le volume"

into 

"volume_slider_aria_label": "Contrôle du volume",
"volume_slider_aria_description": "utiliser les flèches pour contrôler le volume"

And then use the separate values for sliderAriaLabel id and sliderAriaDescription id, which is simple enough in this case. However, in other languages the split into two parts might be more complicated, so we should confirm this with product before making the change.

SivanA-Kaltura (changes requested):

src/components/volume/volume.tsx
valentinStoicaK: I've talked to Sagi and I will open a ticket for translations. In the meantime we can push with english only. I've also added the translation

SivanA-Kaltura (approved):

PR #939, opened 2024-09-11
fix(ADA-1621): bring captions in front of entry title/description

valentinStoicaK:

Used z-index to bring captions in front of Title and Description for audio entries.

**Issue:**

**Fix:**


valentinStoicaK:
I'm going to try another approach

PR #941, opened 2024-09-15
fix(FEC-14080): Accessibility fix for the top bar and the bottom bar

SivanA-Kaltura:

Add bar gradients
Handle bar gradients for native captions in safari

Resolves FEC-14080

### DIFF
--- a/src/components/bottom-bar/_bottom-bar.scss
+++ b/src/components/bottom-bar/_bottom-bar.scss
@@ -1,3 +1,25 @@
+@mixin bottom-bar-gradient {
+  position: absolute;
+  bottom: 0;
+  left: 0;
+  width: 100%;
+  height: var(--playkit-bar-gradient-height);
+  max-height: 200px;
+  opacity: 1;
+  background: linear-gradient(180deg, rgba(0, 0, 0, 0) 0%, rgba(0, 0, 0, 0.8) 100%);
+  content: "";
+}
+
+.player {
+  :global(.#{$video-filter}) {
+    --playkit-bar-gradient-height: 50%;
+  }
+ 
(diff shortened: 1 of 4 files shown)


src/components/bottom-bar/_bottom-bar.scss
```diff
@@ -1,3 +1,25 @@
+@mixin bottom-bar-gradient {
```
lianbenjamin, line 1: just sharing my thoughts:
1. if i understand correctly, `bottom-bar-gradient` is the same gradient being used both in `bottom-bar` and `top-bar`. if this is the case, should we give it a different name and not limit it to `bottom-bar`?
2. also, if we're using it in both places, should we declare the `mixin` in a general place, rather than in `bottom-bar`?

src/components/top-bar/_top-bar.scss
```diff
@@ -73,6 +73,15 @@
 .player.ad-break,
 .player.menu-active {
   &:not(.overlay-active) {
+    :global(.#{$video-filter}) {
+      &:before {
+        @include bottom-bar-gradient();
```
lianbenjamin, line 78: for my understanding, how does it recognize `bottom-bar-gradient` in top-bar if it is not being imported?

lianbenjamin (changes requested):

src/components/top-bar/_top-bar.scss
SivanA-Kaltura: i didn't actually check after just confirming that it works, but it probably works because both of them are later imported in style.scss, and/or because during runtime they're both part of one css document

lianbenjamin (approved):

PR #942, opened 2024-09-15
fix(ADA-1593): shift focus to seekbar when playing/pausing media within small width players

Lkaltura:
issue:
when trying to play/pause a video in a small /extra small /tiny width player using voiceover on mac, the play/pause button disappears after a couple of seconds and the user is unable to pause or play the video again.

fix:
since play/pause button is not in the bottom bar in these widths, switch the focus on the seekbar, which will enable the play/pause functionality.

width <= 480:
image

width > 480:
image

### DIFF
--- a/src/components/play-pause/play-pause.tsx
+++ b/src/components/play-pause/play-pause.tsx
@@ -15,7 +15,7 @@ import {Button} from '../button';
 import {actions as shellActions} from '../../reducers/shell';
 import {ButtonControl} from '../button-control';
 import {withEventManager} from '../../event';
-import {PLAYER_BREAK_POINTS} from "../shell";
+import {PLAYER_SIZE} from '../shell';
 
 /**
  * mapping state to props
@@ -27,7 +27,7 @@ const mapStateToProps = state => ({
   isPlaying: state.engine.isPlaying,
   adBreak: state.engine.adBreak,
   isPlaybackEnded: state.engine.isPlaybackEnded,
-  guiClientRect: state.shell.guiClientRect
+  playerSize: state.shell.playerSize
 });
 
 const COMPONENT_NAME = 'PlayPause';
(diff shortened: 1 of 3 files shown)


src/components/seekbar/seekbar.tsx
```diff
@@ -109,6 +111,13 @@ class SeekBar extends Component<any, any> {
         this.setState({resizing: false});
       }, Number(style.defaultTransitionTime));
     });
+    if (this.props.guiClientRect.width <= PLAYER_BREAK_POINTS.SMALL) {
```
SivanA-Kaltura, line 114: you shouldn't need to check the dimensions in a component explicitly, the active breakpoint should already be updated in state.shell.playerSize

also there's a state called shell.isMobile which might be relevant here as well, depending on what behavior we want on non-tiny mobile screens

src/components/play-pause/play-pause.tsx
```diff
@@ -57,12 +59,14 @@ class PlayPause extends Component<any, any> {
    * @memberof PlayPause
    */
   componentDidMount(): void {
-    const {eventManager, player} = this.props;
-    eventManager.listenOnce(player, player.Event.UI.USER_CLICKED_PLAY, () => {
-      eventManager.listenOnce(player, player.Event.Core.FIRST_PLAY, () => {
-        this._playPauseButtonRef?.focus();
+    const {eventManager, player, guiClientRect} = this.props;
```
SivanA-Kaltura, line 62: please double check that this works in playlist playback, since componentDidMount, theoretically, is called once per component

since this code already existed, i assume it works (and the component is destroyed and recreated for every new entry loaded), but please make sure

SivanA-Kaltura (changes requested):

src/components/play-pause/play-pause.tsx
Lkaltura: originally with voiceOver and playlists, after loading a new entry, the focus shifted back to the whole window.
my fix is not affecting this behavior, meaning the focus still shifts back to the window after loading a new entry.
but I think this is another behavior that is not related to the play/pause functionality and this issue specifically.
let me know if you agree or not.

SivanA-Kaltura (approved):

PR #948, opened 2024-09-24
fix(FEC-14156): quiz overlay behind cvaa freezes

lianbenjamin:

bugfix

**Issue:**
when opening the cvaa (advanced captions settings) while the video is playing and the video reaches a quiz question cue point, then closing the cvaa overlay- the quiz question overlay stays on the player and there is no way to close it.
the root cause is that closing the cvaa causes the removal of overlay-active class, which causes the odd behavior of the quiz question overlay.

**Fix:**
- do not remove the `overlay-active` class if `overlayPortal` component has multiple children- remove it only if there's a single child
- create a reusable `getOverlayPortalElement` function which returns the `overlayPortal` element
- use `getOverlayPortalElement` function in other components
- expose `getOverlayPortalElement` function for a future use outside of core scope

### DIFF
--- a/src/components/overlay/overlay.tsx
+++ b/src/components/overlay/overlay.tsx
@@ -7,6 +7,8 @@ import {actions as overlayActions} from '../../reducers/overlay';
 import {actions as shellActions} from '../../reducers/shell';
 import {default as Icon, IconType} from '../icon';
 import {KeyMap} from '../../utils';
+import {withPlayer} from '../player';
+import {getOverlayPortalElement} from '../overlay-portal';
 
 const COMPONENT_NAME = 'Overlay';
 
@@ -31,6 +33,7 @@ const mapStateToProps = state => ({
  * @extends {Component}
  */
 @connect(mapStateToProps, bindActions({...shellActions, ...overlayActions}))
+@withPlayer
 class Overlay extends Component<any, any> {
   _timeoutId: number | null = null;
   overlayRef: RefObject<HTMLDivElement> = createRef<HTMLDivElement>();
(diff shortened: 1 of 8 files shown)


src/components/overlay/overlay.tsx
```diff
@@ -55,7 +56,10 @@ class Overlay extends Component<any, any> {
       clearTimeout(this._timeoutId);
       this._timeoutId = null;
     }
-    this.props.removePlayerClass(style.overlayActive);
+    // Remove the overlay-active class only when there is a single child
+    if (this.overlayRef.current?.childElementCount === 1) {
```
SivanA-Kaltura, line 60: i don't like this condition because it's for a specific use case and because i'd rather take this information from somewhere else instead of checking the children, if possible.
why, when the overlay is unmounting, is something is still visible on the player ? how does that happen ?

SivanA-Kaltura (changes requested):

src/components/overlay/overlay.tsx
lianbenjamin: as i mentioned in the description, it happens when you open the cvaa while the video is playing and it reaches a quiz question cp. so under the cvaa you have the question overlay. then, when you close the cvaa overlay, the overlay component is unmounted and is removing the overlay-active class which causes the current issue.

src/components/overlay/overlay.tsx
```diff
@@ -141,7 +149,7 @@ class Overlay extends Component<any, any> {
     }
 
     return (
-      <div tabIndex={-1} className={overlayClass.join(' ')} role="dialog" onKeyDown={this.onKeyDown} {...ariaProps}>
+      <div tabIndex={-1} className={overlayClass.join(' ')} role="dialog" onKeyDown={this.onKeyDown} {...ariaProps} ref={this.overlayRef}>
```
SivanA-Kaltura, line 152: do you still need the ref ?

SivanA-Kaltura (changes requested):

src/components/overlay/overlay.tsx
lianbenjamin: forgot to remove it, thanks.

SivanA-Kaltura (approved):

PR #949, opened 2024-09-26
fix(ADA-1593): shift focus to seekbar when playing/pausing media within small width players

Lkaltura:
adjusting the 'playerSize check' order to be executed after the event listener for first_play
and referring directly to this.props.playerSize instead of destructuring it from the props, as its value changes after the component is mounted, and we require the updated size.


SivanA-Kaltura:
@Lkaltura Please fix the conflicts with the master, update the title (the title is not that important, but just to make sure that the squashed commit name will be in the correct format). Also please add a description in the future because it helps understand the scope of the change.

SivanA-Kaltura (approved):

PR #955, opened 2024-11-10
fix(ADA-1791): [ORS] - Player - Screen reader (NVDA) is interpreting time values as clock times

Tzipi-kaltura:

**Issue:**
Times on the seekbar are not clear and can be interpreted as clock times for screen reader users

**Fix:**
Adding to the aria-valuetext hours/minutes/seconds after each time unit.

#### Resolves [ADA-1791](https://kaltura.atlassian.net/browse/ADA-1791)

[ADA-1791]: https://kaltura.atlassian.net/browse/ADA-1791?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/utils/index.ts
```diff
@@ -4,3 +4,4 @@ export {KeyMap, KeyCode} from './key-map';
 export {default as getLogger} from './logger';
 export {withKeyboardA11y} from './popup-keyboard-accessibility';
 export * from './time-format';
+export {getTimeInText} from './duration-humanizer';
```
lianbenjamin, line 7: please also export it on the ui so we can use it from other places. see example here- https://github.com/kaltura/playkit-js-ui/blob/master/src/index.ts#L20, https://github.com/kaltura/playkit-js-ui/blob/master/src/index.ts#L24

lianbenjamin (changes requested):

src/utils/index.ts
Tzipi-kaltura: done

src/utils/duration-humanizer.ts
```diff
@@ -0,0 +1,40 @@
+import * as humanizeDuration from 'humanize-duration';
+
+const getTimeInText = (seconds: number, uiConfig: any) => {
+  try {
+    const durationHumanizer = getDurationHumanizer(uiConfig);
+    if (durationHumanizer) {
+      try {
+        return seconds ? durationHumanizer(seconds * 1000, {round: true}) : '0';
```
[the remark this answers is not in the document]
Tzipi-kaltura: I added it because we don't want to see the numbers after the decimal point, only the integer number same as we see in the time on the bottom bar. but you are right regarding other places so i changed it to be an optional argument.

lianbenjamin (approved):
LGTM, mind the lint errors.

lianbenjamin (approved):

src/utils/index.ts
SivanA-Kaltura, line 7: I think we should be very clear in the name of the method so it would be intuitive what it does.

Time => i think it should be "duration". because we're not handling timestamps, just duration in seconds.
InText => I think it should be "as text"

So I would prefer something like "getDurationAsText" more.

src/utils/duration-humanizer.ts
```diff
@@ -0,0 +1,40 @@
+import * as humanizeDuration from 'humanize-duration';
+
+const getTimeInText = (seconds: number, uiConfig: any, round: boolean = false): string | undefined => {
```
SivanA-Kaltura, line 3: is this copied from the previous implementation that used humanizeDuration or is something unique in this method compared to the previous implementation ?

src/utils/index.ts
```diff
@@ -4,3 +4,4 @@ export {KeyMap, KeyCode} from './key-map';
 export {default as getLogger} from './logger';
 export {withKeyboardA11y} from './popup-keyboard-accessibility';
```
SivanA-Kaltura, line 5: is converting the other places in the player that use humanizeDuration to use this method instead part of this ticket ? If not, we should have a separate ticket for this

src/utils/duration-humanizer.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (durationHumanizer) {
+      try {
+        return seconds ? durationHumanizer(seconds * 1000, {round: round}) : '0';
+      } catch (e: any) {
+        return `${seconds}`;
+      }
+    }
+  } catch (e: any) {
+    // eslint-disable-next-line no-console
+    console.log(e);
```
SivanA-Kaltura, line 15: If logging is necessary here, use logger, not console

src/components/seekbar/seekbar.tsx
```diff
@@ -576,7 +580,7 @@ class SeekBar extends Component<any, any> {
         aria-valuemin={0}
         aria-valuemax={Math.round(this.props.duration)}
         aria-valuenow={Math.round(this.props.currentTime)}
-        aria-valuetext={`${toHHMMSS(this.props.currentTime)} of ${toHHMMSS(this.props.duration)}`}
+        aria-valuetext={`${getTimeInText(props.currentTime, props.player.config.ui, true)} of ${getTimeInText(props.duration, props.player.config.ui, true)}`}
```
SivanA-Kaltura, line 583: if you're only using locale from ui, pass just the locale, not the entire object. it's more readable.

SivanA-Kaltura (changes requested):

src/utils/index.ts
Tzipi-kaltura: this is not part of the ticket. i can create a FEC ticket for that.

src/utils/duration-humanizer.ts
Tzipi-kaltura: it's the same as Lian implemented, i just added round argument (boolean) so you can sent it to humanize-duration api and it will round the result to integer number.

src/components/seekbar/seekbar.tsx
Tzipi-kaltura: done

src/utils/index.ts
Tzipi-kaltura: done

src/utils/duration-humanizer.ts
Tzipi-kaltura: done

SivanA-Kaltura (approved):

PR #959, opened 2024-11-14
fix(ADA-1593): instead of forwardRef , manually set the button reference in the button component

Lkaltura:
issue:
forwardRef didn't pass the ref properly. thus in play-pause component the ref call back when rendering the button component wasn't executed, and so the focus didn't execute as the button was undefined.

solution:
after discussing it with @semarche-kaltura , it was decided to set the reference manually.
since the reference can be a function or an object, I've added two cases to handle that, and adjust the compoents accordingly to send the right props.
now the focus works as expected.

### DIFF
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -20,7 +20,7 @@ const Button = (props) => {
     if (ref && 'current' in ref && buttonRef.current) {
       ref.current = buttonRef.current;
     }
-  }, [setRef, ref]);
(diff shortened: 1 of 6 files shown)


Lkaltura:
after discussing with @SivanA-Kaltura 
it's preferred to keep the forwardRef in the meantime, investigate it's functionality thoroughly first.
to address the focus issue, refer to this PR: https://github.com/kaltura/playkit-js-ui/pull/962

PR #962, opened 2024-11-25
fix(ADA-1593): fix focus issue on play-pause button with different player sizes

Lkaltura:
issue:
play-pause button is undefined , so we weren't able to perform focus action on the button

solution:
after discussing with @SivanA-Kaltura 
it's better to change the way we reference play-pause button to be able to focus on the button element when it's on bigger player sizes.

### DIFF
--- a/src/components/play-pause/play-pause.tsx
+++ b/src/components/play-pause/play-pause.tsx
@@ -97,8 +97,9 @@ class PlayPause extends Component<any, any> {
     const playbackStateText = this.props.isPlayingAdOrPlayback ? this.props.pauseText : this.props.playText;
     const labelText = isStartOver ? this.props.startOverText : playbackStateText;
     return (
-      <div ref={this._playPauseButtonRef} style={{ display: 'contents' }}>
-        <ButtonControl name={COMPONENT_NAME}>
+
+      <ButtonControl name={COMPONENT_NAME}>
+        <div ref={this._playPauseButtonRef} >
           <Tooltip label={labelText}>
             <Button
               tabIndex="0"
@@ -116,9 +117,11 @@ class PlayPause extends Component<any, any> {
               )}
             </Button>
           </Tooltip>
-        </ButtonControl>
-      </div>
-    );
+        </div>
+      </ButtonControl>
+
+  )
+    ;
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/components/play-pause/play-pause.tsx
```diff
@@ -65,7 +65,7 @@ class PlayPause extends Component<any, any> {
     eventManager.listenOnce(player, player.Event.UI.USER_CLICKED_PLAY, () => {
       eventManager.listenOnce(player, player.Event.Core.FIRST_PLAY, () => {
         if (!smallSizes.includes(this.props.playerSize)) {
-          this._playPauseButtonRef?.focus();
+          this._playPauseButtonRef?.current?.querySelector("button")?.focus();
```
SivanA-Kaltura, line 68: this focuses on the correct play button (middle button / bottom bar button) ?
looks good

my only comment would be to consider maybe using [focusElement](https://github.com/kaltura/playkit-js-ui/blob/62cb9bc1cb0975ca3c0ef8a7eb735c9d95392e5b/src/utils/focus-element.ts#L9) if you think that there can be a scenario where this code would be called when the button isn't visible yet

SivanA-Kaltura (changes requested):

src/components/play-pause/play-pause.tsx
Lkaltura: answers to your questions:
1. yes, on first play: 
Player is on smaller sizes: as before, focus on seekbar, no changes here
player on bigger sizes: focus on bottom bar play-pause button :
image

2. Before first play When side panel is open -> click play: focus is on seekbar
Before first play When side panel is closed-> click play: focus is on bottom Play-pause button, opening side panel after first play shouldn't automatically shift focus to play-pause button as the focus would be on side panel element

3. The two play button you're referring to is the one in the middle and the one at the bottom bar,
the event listener (eventManager.listenOnce) inside play-pause was there before my changes, and it behaves the same as it was before. It cannot be called a second time after first play. 


* please pay attention i added another commit:
 added style={{ display: 'contents' } to make the <div> invisible to the layout to ensure it's unaffected. because without it, the <div> hid the rewind forward buttons:
image
 after : 
image

Let me know if these answer your questions.

SivanA-Kaltura: @Lkaltura 
Regarding 2, if both event listeners (in the 2 buttons) are called, but only one of them actually does something, that's fine. I was asking if they don't both try to focus on their inner play button at the same time or something like that.

Regarding the change in the style, not sure i understand why that was needed. Please try to explain again in text, if that doesn't work let's try in voice.

Lkaltura: I messaged you

SivanA-Kaltura (approved):

PR #968, opened 2024-12-18
fix(ADA-1751): [TR] ADA Loading play not announced

Tzipi-kaltura:

**Issue:**
Screen reader user doesn't have indication on loading state.

**Fix:**
Adding hidden element that visible only to screen reader. in case loading take more than second, "loading" message will be announced.

### DIFF
--- a/src/components/spinner/_spinner.scss
+++ b/src/components/spinner/_spinner.scss
@@ -47,3 +47,11 @@
     }
   }
 }
+.sr-only {
+  position: absolute;
+  left: -10000px;
+  top: auto;
+  width: 1px;
+  height: 1px;
+  overflow: hidden;
+}
--- a/src/components/spinner/spinner.tsx
+++ b/src/components/spinner/spinner.tsx
@@ -1,7 +1,12 @@
 import style from '../../styles/style.scss';
-import {h, Component, VNode} from 'preact';
+import {h, Component, VNode, RefObject, createRef} from 'preact';
+import {Text, withText} from 'preact-i18n';
 
 const COMPONENT_NAME = 'Spinner';
+
+const translates = {
+  loading: <Text id="spinner.loading">Loading</Text>
+};
 
 /**
  * Spinner component
@@ -10,7 +15,18 @@ const COMPONENT_NAME = 'Spinner';
  * @example <Spinner />
  * @extends {Component}
  */
+@withText(translates)
 class Spinner extends Component<any, any> {
+  loadingSpan: RefObject<HTMLElement> = createRef<HTMLElement>();
+
+  componentDidMount() {
+    setTimeout(() => {
+      if (this.loadingSpan.current) {
+        this.loadingSpan.current.innerText = this.props.loading;
+      }
+    }, 1000);
+  }
+
   /**
    * render component
    *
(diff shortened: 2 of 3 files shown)


translations/en.i18n.json
```diff
@@ -140,6 +140,9 @@
     },
     "watermark": {
       "watermark_alt_text": "Watermark"
+    },
+    "spinner": {
+      "loading": "Loading"
```
SivanA-Kaltura, line 145: is there a ticket for translations ?

src/components/spinner/spinner.tsx
```diff
@@ -1,16 +1,32 @@
...
  * @class Spinner
  * @example <Spinner />
  * @extends {Component}
  */
+@withText(translates)
 class Spinner extends Component<any, any> {
+  loadingSpan: RefObject<HTMLElement> = createRef<HTMLElement>();
+
+  componentDidMount() {
+    setTimeout(() => {
```
SivanA-Kaltura, line 23: what does the setTimeout do ?

src/components/spinner/_spinner.scss
```diff
@@ -47,3 +47,11 @@
     }
   }
 }
+.sr-only {
```
SivanA-Kaltura, line 50: what does this do ?

SivanA-Kaltura (changes requested):

translations/en.i18n.json
Tzipi-kaltura: yes

src/components/spinner/spinner.tsx
Tzipi-kaltura: only if the loading appears more than one second screen reader will announce: "loading" (asked by product)

src/components/spinner/_spinner.scss
Tzipi-kaltura: it's screen reader only element so need to hide it from visibility. this is common css for screen reader only elements (kms also use it)

src/components/spinner/spinner.tsx
SivanA-Kaltura: two things
- you need to start the timer from the moment the loading is VISIBLE, not mounted
- i didn't understand the use of ref here. is using a boolean not good enough ?

SivanA-Kaltura (changes requested):

src/components/spinner/spinner.tsx
Tzipi-kaltura: 1. I checked it and when the setTimeout is started, loading is already visible.
2. need to update the span with the "loading" message if we want the screen reader to announce it immediately. in order to do it need to save the span in ref and update it.

SivanA-Kaltura: - i would still add a visibility check. i don't know if you can count on it.
- what happens if you don't use a ref ?

SivanA-Kaltura (changes requested):

src/components/spinner/spinner.tsx
Tzipi-kaltura: visibility check was added.

SivanA-Kaltura (approved):

PR #976, opened 2025-01-22
fix(SUP-46391): Player Live Full Screen Bug when browser window is not maximized

yambenari24:
Issue: 
When browser width is small, side pannel is on, and enter full screen, there is a gap between side pannel and player.

Fix:
Fix timing issue : the issue related to the timing of the DOM updates. When resizing the window or triggering an actions that affect the layout, the client rect of the player container getBoundingClientRect() is not updated immediately after the resize. This causes discrepancies between the actual layout and the values being used in the state.

solved [SUP-46391](https://kaltura.atlassian.net/browse/SUP-46391)

[SUP-46391]: https://kaltura.atlassian.net/browse/SUP-46391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/shell/shell.tsx
+++ b/src/components/shell/shell.tsx
@@ -43,6 +43,7 @@ const mapStateToProps = state => ({
 });
 
 const ON_PLAYER_RECT_CHANGE_DEBOUNCE_DELAY: number = 100;
+const RESIZE_DEBOUNCE_DELAY: number = 0;
 
 const PLAYER_SIZE: {[size: string]: string} = {
   TINY: 'tiny',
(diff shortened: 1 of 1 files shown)


src/components/shell/shell.tsx
```diff
@@ -212,7 +213,7 @@ class Shell extends Component<any, any> {
    */
   componentDidMount() {
     const {player, eventManager} = this.props;
-    eventManager.listen(window, 'resize', debounce(this._onWindowResize, ON_PLAYER_RECT_CHANGE_DEBOUNCE_DELAY));
+    eventManager.listen(window, 'resize', debounce(this._onWindowResize, RESIZE_DEBOUNCE_DELAY));
```
SivanA-Kaltura, line 216: this fix is only for fullscreen, right ?

SivanA-Kaltura (changes requested):

src/components/shell/shell.tsx
MosheMaorKaltura: Only full screen

yambenari24: yes @SivanA-Kaltura

MosheMaorKaltura (approved):

PR #978, opened 2025-02-01
fix(FEC-14289): Player v7 | Player doesn't recover well after network disconnection

roeedean:
After player had network disconnection error and user clicked on 'Try Again', player should be **destroyed** and we should **setup** a new player before using **loadMedia**.

Related PR: https://github.com/kaltura/kaltura-player-js/pull/914

Resolves [FEC-14289](https://kaltura.atlassian.net/browse/FEC-14289)

[FEC-14289]: https://kaltura.atlassian.net/browse/FEC-14289

### DIFF
--- a/src/components/error-overlay/error-overlay.tsx
+++ b/src/components/error-overlay/error-overlay.tsx
@@ -59,6 +59,16 @@ class ErrorOverlay extends Component<any, any> {
    */
   handleClick = (): void => {
     const mediaInfo = this.props.player.getMediaInfo();
+    const config = this.props.player.config;
+
+    // Destroy the existing player
+    this.props.player.destroy();
+    this.props.player = null;
+
+    // Setup player again
+    this.props.player = KalturaPlayer.setup(config);
+
+    // Load the media again
     this.props.player.loadMedia(mediaInfo);
   };
 
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -33,6 +33,9 @@
     "outDir": "./lib/",
     "allowSyntheticDefaultImports": true
   },
-  "files": ["./src/types/kaltura-player-js.d.ts"],
+  "files": [
+    "./src/types/kaltura-player-js.d.ts",
+    "node_modules/@playkit-js/kaltura-player-js/ts-typed/kaltura-player.d.ts"
+  ],
   "include": ["src/**/*"]
 }


src/components/error-overlay/error-overlay.tsx
```diff
@@ -59,6 +59,16 @@ class ErrorOverlay extends Component<any, any> {
    */
   handleClick = (): void => {
     const mediaInfo = this.props.player.getMediaInfo();
+    const config = this.props.player.config;
+
+    // Destroy the existing player
+    this.props.player.destroy();
+    this.props.player = null;
```
SivanA-Kaltura, line 66: does calling player destroy also destroy/unmount this component ?

SivanA-Kaltura (changes requested):

src/components/error-overlay/error-overlay.tsx
```diff
@@ -59,6 +59,16 @@ class ErrorOverlay extends Component<any, any> {
    */
   handleClick = (): void => {
     const mediaInfo = this.props.player.getMediaInfo();
+    const config = this.props.player.config;
```
SivanA-Kaltura, line 62: the question is whether this is identical to the config that this player was created with.

src/components/error-overlay/error-overlay.tsx
```diff
@@ -59,6 +59,16 @@ class ErrorOverlay extends Component<any, any> {
...
+
+    // Destroy the existing player
+    this.props.player.destroy();
```
SivanA-Kaltura, line 65: what happens if something has a reference to the old player ? like dual screen for example ?

SivanA-Kaltura (changes requested):

src/components/error-overlay/error-overlay.tsx
roeedean: the config is identical.

src/components/error-overlay/error-overlay.tsx
roeedean: it destroys only the "kaltura-player-container" inside the component of the "player-placeholder", so there will be no "kaltura-player-container" inside the component.
after we are doing player.setup again it will recreate the same player ("kaltura-player-container") that was there before, and the new player will have the same reference as the old player.

src/components/error-overlay/error-overlay.tsx
roeedean: it destroys the player ("kaltura-player-container") inside the component of the "player-placeholder".

MosheMaorKaltura (approved):

PR #980, opened 2025-02-07
fix(SUP-46685): Multiple Chapters in Quiz Break the Timeline

yambenari24:
**Issue:**
Timing issue: in some cases the delay in the setTimeOut in SeekBar.tsx file is not enough, and the bottomBarClientRect in BottomBar.tsx being updated too late, just after the set time out scope is being executed. This affects the seekBar width, which creates a problem with position the cue-points on the seekBar.

**Fix:**
Add an event that will update the seekBar every time the bottomBarClientRect is being updated. The seekBar.tsx will always get the most updated width.

solved-[SUP-46685](https://kaltura.atlassian.net/browse/SUP-46685)

[SUP-46685]: https://kaltura.atlassian.net/browse/SUP-46685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/bottom-bar/bottom-bar.tsx
+++ b/src/components/bottom-bar/bottom-bar.tsx
@@ -10,6 +10,7 @@ import {withEventManager} from '../../event';
 import {withPlayer} from '../player';
 import {filterControlsByPriority} from './bottom-bar-utils';
 import {BottomBarRegistryManager, bottomBarRegistryManager} from './bottom-bar-registry-manager';
+import {BottomBarClientRectEvent} from '../../event/events/bottom-bar-client-rect-event';
 
 const LOWER_PRIORITY_CONTROLS: string[][] = [
   ['PictureInPicture'],
@@ -99,7 +100,8 @@ class BottomBar extends Component<any, any> {
     const currentControlsWidth = this._getControlsWidth(resizeObserverEntry.target as HTMLElement);
     this._maxControlsWidth = Math.max(this._maxControlsWidth, currentControlsWidth);
     if (barWidth !== this.currentBarWidth) {
-      document.dispatchEvent(new Event('bottomBarClientRectChanged'));
+      const {player} = this.props;
+      player.dispatchEvent(new BottomBarClientRectEvent());
       this.currentBarWidth = barWidth;
       const currCrlWidth = this.props.guiClientRect.width <= PLAYER_BREAK_POINTS.SMALL ? CRL_WIDTH + CRL_MARGIN / 2 : CRL_WIDTH + CRL_MARGIN;
       const lowerPriorityControls = LOWER_PRIORITY_CONTROLS.filter(c => this.state.activeControls[c[0]]);
--- a/src/components/bottom-bar/bottom-bar.tsx
+++ b/src/components/bottom-bar/bottom-bar.tsx
@@ -99,6 +99,7 @@ class BottomBar extends Component<any, any> {
     const currentControlsWidth = this._getControlsWidth(resizeObserverEntry.target as HTMLElement);
     this._maxControlsWidth = Math.max(this._maxControlsWidth, currentControlsWidth);
     if (barWidth !== this.currentBarWidth) {
+      window.dispatchEvent(new Event('bottomBarClientRectUpdated'));
       this.currentBarWidth = barWidth;
       const currCrlWidth = this.props.guiClientRect.width <= PLAYER_BREAK_POINTS.SMALL ? CRL_WIDTH + CRL_MARGIN / 2 : CRL_WIDTH + CRL_MARGIN;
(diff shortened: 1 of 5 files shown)


src/components/bottom-bar/bottom-bar.tsx
```diff
@@ -99,6 +99,7 @@ class BottomBar extends Component<any, any> {
     const currentControlsWidth = this._getControlsWidth(resizeObserverEntry.target as HTMLElement);
     this._maxControlsWidth = Math.max(this._maxControlsWidth, currentControlsWidth);
     if (barWidth !== this.currentBarWidth) {
+      window.dispatchEvent(new Event('bottomBarClientRectUpdated'));
```
semarche-kaltura, line 102: `@withEventManager` decorator adds eventManager into component that you can use instead of dispatching events directly to window. The issue here is next: if on page you have more than one Kaltura player events gonna be handled by every of them instead of specific player, where bottomBar size changed

src/components/seekbar/seekbar.tsx
```diff
@@ -98,6 +98,11 @@ class SeekBar extends Component<any, any> {
     }
   ];
 
+  updateSeekBarClientRect = () => {
```
semarche-kaltura, line 101: name of method similar with method from props, I would recommend rename to `handleUpdateSeekBarClientRect`, etc to simplify code reading

src/components/seekbar/seekbar.tsx
```diff
@@ -138,6 +138,7 @@ class SeekBar extends Component<any, any> {
   componentWillUnmount(): void {
     document.removeEventListener('mouseup', this.onPlayerMouseUp);
     document.removeEventListener('mousemove', this.onPlayerMouseMove);
+    this.props.eventManager.unlisten(document, 'bottomBarClientRectChanged', this.handleUpdateSeekBarClientRect);
```
SivanA-Kaltura, line 141: actually, and you can double check, when you use withEventManager, it [should remove the event listeners by itself on unmount](https://github.com/kaltura/playkit-js-ui/blob/fa0edb505f9230b953011a916ce090e6f0cc09f6/src/event/with-event-manager.tsx#L22), without you having to do it in your code

also, when you export the event name to EventType, please use EventType.[new event name] here instead of writing the string. it will just make refactor easier if the string can be changed in one place and affect every place that uses the event.

also that way we can reuse this event in plugins if we need to, by just importing EventType in that plugin.

src/components/seekbar/seekbar.tsx
```diff
@@ -124,6 +122,8 @@ class SeekBar extends Component<any, any> {
         }
       });
     });
+
+    this.props.eventManager.listen(document, 'bottomBarClientRectChanged', () => this.handleUpdateSeekBarClientRect());
```
SivanA-Kaltura, line 126: see my comment below about the EventType

src/components/seekbar/seekbar.tsx
```diff
@@ -108,14 +113,7 @@ class SeekBar extends Component<any, any> {
...
     this.props.updateSeekbarClientRect(clientRect);
-    eventManager.listen(player, EventType.GUI_RESIZE, () => {
-      this.setState({resizing: true});
-      setTimeout(() => {
-        const clientRect = this._seekBarElement.getBoundingClientRect();
-        this.props.updateSeekbarClientRect(clientRect);
-        this.setState({resizing: false});
-      }, Number(style.defaultTransitionTime));
-    });
+
```
SivanA-Kaltura, line 116: please do a blame on the setTimeout line, check which ticket it was fixed for, and try to see if your changes did not cause a regression for that issue.
I don't think they did, but let's make sure

src/components/bottom-bar/bottom-bar.tsx
```diff
@@ -99,6 +99,7 @@ class BottomBar extends Component<any, any> {
     const currentControlsWidth = this._getControlsWidth(resizeObserverEntry.target as HTMLElement);
     this._maxControlsWidth = Math.max(this._maxControlsWidth, currentControlsWidth);
     if (barWidth !== this.currentBarWidth) {
+      document.dispatchEvent(new Event('bottomBarClientRectChanged'));
```
SivanA-Kaltura, line 102: looks good !
lets make it a bit more organized
- make it a FakeEvent like GuiResizeEvent
- add it to EventType
- fire it on the player and not the document

SivanA-Kaltura (changes requested):

src/components/seekbar/seekbar.tsx
yambenari24: Looks like it was part of a product requirement:
https://kaltura.atlassian.net/browse/FEC-10311

SivanA-Kaltura (approved):

PR #984, opened 2025-02-21
fix(ADA-1745) - Create alternative aria-label for close button 

valentinStoicaK:
This solves https://kaltura.atlassian.net/browse/ADA-1745
Related to  https://github.com/kaltura/playkit-js-info/pull/112 and https://github.com/kaltura/playkit-js-downloads/pull/84

**Issue:**

**Fix:**

### DIFF
--- a/src/components/overlay/overlay.tsx
+++ b/src/components/overlay/overlay.tsx
@@ -56,6 +56,7 @@ const mapStateToProps = state => ({
 class Overlay extends Component<OverlayProps, any> {
   private _timeoutId: NodeJS.Timeout | null = null;
   private _wasPlayed = false; // keep state of the player so we can resume if needed
+  static defaultProps = {closeAriaLabel: (<Text id="overlay.close" />) as unknown as string};
 
   /**
    * componentWillMount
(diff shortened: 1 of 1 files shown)


src/components/overlay/overlay.tsx
```diff
@@ -155,7 +155,10 @@ class Overlay extends Component<OverlayProps, any> {
               props.onClose();
             }}
             onKeyDown={this.onCloseButtonKeyDown}
-            aria-label={(<Text id="overlay.close" />) as unknown as string}
+            aria-label={
+              props.closeAriaLabel ||
```
SivanA-Kaltura, line 159: where is this value being set ?

valentinStoicaK: closeAriaLabel is an optional prop that is set when needed. For now I am setting it in Video Info and Downloads.

semarche-kaltura: I would recommend use `(<Text id="overlay.close" />) as unknown as string` as default property of `closeAriaLabel`. Then you can use `aria-label={props.closeAriaLabel}`

semarche-kaltura (changes requested):

semarche-kaltura (approved):

SivanA-Kaltura (approved):

PR #986, opened 2025-02-24
fix (SUP-46685): Multiple Chapters in Quiz Break the Timeline

yambenari24:
**Issue:**
The question appears a few seconds earlier than expected on the timeline.
Root Cause:
The issue stems from the player hover state not being updated. As a result:
The bottom bar doesn’t adjust its width correctly.
This prevents the seek bar from updating its width.
The incorrect seek bar width leads to miscalculations in the question timing.

**Fix:**
To resolve this, an event is dispatched within useLayoutEffect, right before the browser paints the element. This ensures that the seek bar’s dimensions are recalculated correctly, resulting in accurate question timing on the timeline.

The only change in this branch:
1. Add new event
2. Listen to this event
3. In Ivq repo, there is a dispatching for this event

solved [SUP-46685](https://kaltura.atlassian.net/browse/SUP-46685)

[SUP-46685]: https://kaltura.atlassian.net/browse/SUP-46685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/shell/shell.tsx
```diff
@@ -219,6 +219,10 @@ class Shell extends Component<any, any> {
     this._playerResizeWatcher.init(document.getElementById(this.props.targetId)!);
     eventManager.listen(this._playerResizeWatcher, EventType.RESIZE, debounce(this._onWindowResize, ON_PLAYER_RECT_CHANGE_DEBOUNCE_DELAY));
     eventManager.listen(player, player.Event.FIRST_PLAY, () => this._onWindowResize());
+    eventManager.listen(player, EventType.PLAYER_HOVER_STATE_CHANGED, () => {
```
semarche-kaltura, line 222: I would change event name, since `PLAYER_HOVER_STATE_CHANGED` doesn't reflect that player got hover. Here or pass true/false in event payload or change name that indicates that player got hover (and not lost the hover)

yambenari24: i renamed it

semarche-kaltura (approved):

PR #996, opened 2025-03-10
FEC-14416 check if portal element exists before querying it

muli:

**Issue:**
https://kaltura.atlassian.net/browse/FEC-14416

**Fix:**
Make sure that the portal element is not `null` or `undefined` before accessing its `childElementCount` property.

### DIFF
--- a/src/components/overlay/overlay.tsx
+++ b/src/components/overlay/overlay.tsx
@@ -85,7 +85,7 @@ class Overlay extends Component<OverlayProps, any> {
     } else {
       // Remove the overlay-active class only when overlay portal has a single child
       const overlayPortalEl = getOverlayPortalElement(this.props.player);
-      if (overlayPortalEl!.childElementCount <= 1) {
+      if (overlayPortalEl && overlayPortalEl?.childElementCount <= 1) {
         this.props.removePlayerClass!(style.overlayActive);
       }
     }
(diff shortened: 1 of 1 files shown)


muli:
@SivanA-Kaltura, can you please take a look? I'm unable to add reviewers because I only have read only access.

src/components/overlay/overlay.tsx
```diff
@@ -85,7 +85,7 @@ class Overlay extends Component<OverlayProps, any> {
     } else {
       // Remove the overlay-active class only when overlay portal has a single child
       const overlayPortalEl = getOverlayPortalElement(this.props.player);
-      if (overlayPortalEl!.childElementCount <= 1) {
+      if (overlayPortalEl && overlayPortalEl?.childElementCount <= 1) {
```
MosheMaorKaltura, line 88: you don't need to check if overlayPoratlE1 exist anymore, since using ? operator

muli: You're right. Removed.

MosheMaorKaltura (approved):

src/components/overlay/overlay.tsx
```diff
@@ -85,7 +85,7 @@ class Overlay extends Component<OverlayProps, any> {
     } else {
       // Remove the overlay-active class only when overlay portal has a single child
       const overlayPortalEl = getOverlayPortalElement(this.props.player);
-      if (overlayPortalEl!.childElementCount <= 1) {
+      if (overlayPortalEl && overlayPortalEl.childElementCount <= 1) {
```
semarche-kaltura, line 88: with TS you can use `?`
`overlayPortalEl?.childElementCount <= 1`

muli: No, because if either `overlayPortalEl` or `childElementCount` are `undefined`, you then the expression will be resolved as `undefined <= 1`, which isn't allowed in TypeScript.

PR #997, opened 2025-03-10
fix (SUP-47338): Watermark On Player Not Responsive

yambenari24:
**Issue:**
Currently when client applies a watermark image, it is just applied on the player "as is" without calculate the proportion considering the player height and width.

**Fix:**
Calculate the aspect ratio of the image, and apply new height and width based on the player size.
Listen to resize event and update the image size whenever resize is being dispatched.

solved https://kaltura.atlassian.net/browse/SUP-47338

### DIFF
--- a/src/components/watermark/watermark.tsx
+++ b/src/components/watermark/watermark.tsx
@@ -39,8 +39,8 @@ const ENTRY_VAR = '{entryId}';
 class Watermark extends Component<any, any> {
   _timeoutId: number | null = null;
   _imgAspectRatio: number | null = null;
-  _imgWidth: number | null = null;
-  _imgHeight: number | null = null;
+  _originalImgWidth: number | null = null;
+  _originalImgHeight: number | null = null;
   _playerHeight: number | null = null;
   _playerWidth: number | null = null;
   /**
@@ -109,8 +109,8 @@ class Watermark extends Component<any, any> {
     const img = new Image();
     img.src = this.props.config.img;
     img.onload = () => {
-      this._imgWidth = img.naturalWidth;
-      this._imgHeight = img.naturalHeight;
+      this._originalImgWidth = img.naturalWidth;
+      this._originalImgHeight = img.naturalHeight;
       this._imgAspectRatio = img.naturalWidth / img.naturalHeight;
       this.setState({imgWidth: img.naturalWidth, imgHeight: img.naturalHeight});
     };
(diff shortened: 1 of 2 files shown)


src/components/watermark/watermark.tsx
```diff
@@ -90,7 +104,39 @@ class Watermark extends Component<any, any> {
...
-  
+
+  private _loadImageDimension(): void {
+    const img = new Image();
+    img.src = this.props.config.img;
+    img.onload = () => {
+      this._imgWidth = img.naturalWidth;
+      this._imgHeight = img.naturalHeight;
+      this._imgAspectRatio = img.naturalWidth / img.naturalHeight;
+      this.setState({imgWidth: img.naturalWidth, imgHeight: img.naturalHeight});
```
SivanA-Kaltura, line 115: so the image starts out with it's original dimensions, and then gets resized ?

and there's no visible "jump" after the first onPlayerResize ?

src/components/watermark/watermark.tsx
```diff
@@ -90,7 +104,39 @@ class Watermark extends Component<any, any> {
       }
     }
   }
-  
+
+  private _loadImageDimension(): void {
+    const img = new Image();
+    img.src = this.props.config.img;
+    img.onload = () => {
+      this._imgWidth = img.naturalWidth;
```
SivanA-Kaltura, line 112: @this._imgWidth / height stores the ORIGINAL values, so please rename them in order to distinguish them from the values in the state

src/components/watermark/watermark.tsx
```diff
@@ -90,7 +104,39 @@ class Watermark extends Component<any, any> {
...
+    if (playerContainer) {
+      const {width, height} = playerContainer.getBoundingClientRect();
+      this._playerWidth = width;
+      this._playerHeight = height;
+    }
+  }
+
+  private onPlayerResize(): void {
+    const playerContainer = document.getElementById(this.props.targetId);
+    if (!playerContainer || !this._imgAspectRatio || !this._imgWidth || !this._imgHeight) return;
```
SivanA-Kaltura, line 130: if imgWidth and height aren't set, aspect ratio wouldn't be set as well, no ?
so it's enough to check aspect ratio and player container

src/components/watermark/watermark.tsx
```diff
@@ -146,7 +192,16 @@ class Watermark extends Component<any, any> {
       <div className={styleClass.join(' ')}>
         <a href={this.state.urlLink} target="_blank" rel="noopener noreferrer">
           <Localizer>
-            <img src={props.config.img} alt={(<Text id="watermark.watermark_alt_text" />) as unknown as string} />
+            {this.state.imgWidth && this.state.imgHeight && (
+              <img
+                src={props.config.img}
+                alt={(<Text id="watermark.watermark_alt_text" />) as unknown as string}
+                style={{
+                  width: `${this.state.imgWidth}px`,
```
SivanA-Kaltura, line 200: btw, what happens if you keep one of the values as auto ? does it scale automatically ?

SivanA-Kaltura (changes requested):

src/components/watermark/watermark.tsx
yambenari24: @SivanA-Kaltura 
The image always start with its original position and changes its size only when player is resized.
There is no visible jump on the first resize, because I show the image only when i have in my hands the imgWidth and imgHeight

src/components/watermark/watermark.tsx
yambenari24: yes, it's scale automatically and change the size of the image not as expected

src/components/watermark/watermark.tsx
yambenari24: This is check for the typing (validate that image width that i use is not null)

semarche-kaltura (approved):

SivanA-Kaltura (approved):

PR #1001, opened 2025-03-24
fix(SUP-47502): Play/Pause button on images in embed

inbalvasserman:
issue:
on small sizes with image play/pause button can be seen on screen

root- cause
there is css style for the play/pause icon for small sizes

solution:
check if it image, then add class to make css not be implemented

solve [SUP-47502](https://kaltura.atlassian.net/browse/SUP-47502)

[SUP-47502]: https://kaltura.atlassian.net/browse/SUP-47502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/playback-controls/playback-controls.tsx
+++ b/src/components/playback-controls/playback-controls.tsx
@@ -11,7 +11,8 @@ import {PlayerArea} from '../../components/player-area';
  * @returns {Object} - mapped state to this component
  */
 const mapStateToProps = state => ({
-  playlist: state.engine.playlist
+  playlist: state.engine.playlist,
+  image: state.engine.isImg
 });
 
 const COMPONENT_NAME = 'PlaybackControls';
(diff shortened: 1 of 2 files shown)


src/components/playback-controls/playback-controls.tsx
```diff
@@ -60,9 +61,10 @@ class PlaybackControls extends Component<any, any> {
     if (shouldUpdate) {
       this.setState({shouldUpdate: false});
     }
+    const imageClass = this.props.image ? 'playkit-image' : '';
```
SivanA-Kaltura, line 64: isn't the className image and not playkit-image ?
also its more correct to use style.[className], because the actual class name will have a hash added to it

SivanA-Kaltura (changes requested):

SivanA-Kaltura (approved):

PR #1017, opened 2025-04-29
fix (SUP-47965): Hotkey F is not working when the focus is on another button in the video

yambenari24:
**Issue:**
On text box, when navigate to the text box with tab, and press F it opens the fullscreen mode instead write the letter 'f'

**Fix:**
In a case of text area or input or any text box, ignore the hotkey F.

solved [SUP-47965](https://kaltura.atlassian.net/browse/SUP-47965)

[SUP-47965]: https://kaltura.atlassian.net/browse/SUP-47965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/shell/shell.tsx
+++ b/src/components/shell/shell.tsx
@@ -144,12 +144,8 @@ class Shell extends Component<any, any> {
     }
   }
 
-  toggleFullscreen(): void {
-    if (this.props.player.isFullscreen()) {
-      this.props.player.exitFullscreen();
-    } else {
-      this.props.player.enterFullscreen();
-    }
+  toggleFullscreen(): void { 
+    this.props.player.isFullscreen() ? this.props.player.exitFullscreen() : this.props.player.enterFullscreen();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/components/shell/shell.tsx
```diff
@@ -179,14 +179,18 @@ class Shell extends Component<any, any> {
    * @returns {void}
    */
   onKeyDown = (e: KeyboardEvent): void => {
+    const target = e.target as HTMLElement;
+    const isInput = target.tagName === 'INPUT' || target.tagName === 'TEXTAREA' || target.isContentEditable;
```
lianbenjamin, line 183: instead of checking the `tagName`, can we use for example - `target instanceof HTMLInputElement`? (same for `textarea`)

lianbenjamin (changes requested):

src/components/shell/shell.tsx
yambenari24: changed it, thanks!

lianbenjamin (approved):

PR #1022, opened 2025-05-12
fix(ADA-2784): Add transitionToState to onKeyDown

valentinStoicaK:
This solves https://kaltura.atlassian.net/browse/ADA-2784 by adding transitionToState to the onKeyDown

**Issue:**

**Fix:**

### DIFF
--- a/src/components/cvaa-overlay/custom-captions-window.tsx
+++ b/src/components/cvaa-overlay/custom-captions-window.tsx
@@ -52,8 +52,11 @@ class CustomCaptionsWindow extends Component<any, any> {
    * @memberof CustomCaptionsWindow
    */
   onKeyDown = (e: KeyboardEvent): void => {
-    if (e.keyCode === KeyMap.ENTER) {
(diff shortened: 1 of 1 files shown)


src/components/cvaa-overlay/custom-captions-window.tsx
```diff
@@ -1,204 +1,207 @@
...
+
+  /**
+   * on key down handler
+   *
+   * @param {KeyboardEvent} e - keyboard event
+   * @returns {void}
+   * @memberof CustomCaptionsWindow
+   */
+  onKeyDown = (e: KeyboardEvent): void => {
+    if (e.keyCode === KeyMap.ENTER || e.keyCode === KeyMap.SPACE) {
```
SivanA-Kaltura, line 55: Please re apply this fix and then save with "change without formatting"
Otherwise it will be very hard in the future to understand what the change was when looking at the commit because the entire file appears modified

SivanA-Kaltura (changes requested):

SivanA-Kaltura (approved):

PR #1024, opened 2025-05-14
fix(FEC-14483): clipToParent | Upper bar buttons position is off

lianbenjamin:

fixing collision of top-bar's right controls and left controls.

### DIFF
--- a/src/components/top-bar/_top-bar.scss
+++ b/src/components/top-bar/_top-bar.scss
@@ -1,4 +1,6 @@
 .player .top-bar {
+  display: flex;
+  justify-content: space-between;
   background: linear-gradient(180deg, rgba(0, 0, 0, 0.8) 0%, rgba(0, 0, 0, 0) 100%);
   color: #fff;
   visibility: hidden;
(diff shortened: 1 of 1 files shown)


src/components/top-bar/_top-bar.scss
```diff
@@ -50,7 +52,7 @@
   .right-controls {
     float: right;
     text-align: left;
-    margin: #{$top-bar-top-bottom-gutter}px #{$gui-gutter}px #{$top-bar-top-bottom-gutter}px 0;
+    margin: #{$top-bar-top-bottom-gutter}px #{$gui-gutter}px #{$top-bar-top-bottom-gutter}px 32px;
```
SivanA-Kaltura, line 55: what does this margin change do ?

SivanA-Kaltura (changes requested):

src/components/top-bar/_top-bar.scss
lianbenjamin: it adds some space to the left-most edge of the right-controls - it spaces between left and right controls when they collide. so bottom line it makes it look better.

SivanA-Kaltura (approved):

PR #1049, opened 2025-07-07
fix(SUP-48894): Audio flavor selector bug

Tzipi-kaltura:
### Description of the Changes

**Issue:**
In case dropdown under settings (audio, captions) has many items, the most top item is not clickable because top-bar element overlapping it.

**Fix:**
Change the max-height value by subtract the top-bar and the bottom-bar height.

#### Resolves [SUP-48894](https://kaltura.atlassian.net/browse/SUP-48894)

[SUP-48894]: https://kaltura.atlassian.net/browse/SUP-48894?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1061, opened 2025-07-24
fix(SUP-48894): Audio flavor selector bug

Tzipi-kaltura:

continue to - https://github.com/kaltura/playkit-js-ui/pull/1049
**Issue:**
when dropdown bounding is inside topbar element top most element is not clickable 

**Fix:**
check bounding of menu element against the top bar bounding

#### Resolves [SUP-48894](https://kaltura.atlassian.net/browse/SUP-48894)

[SUP-48894]: https://kaltura.atlassian.net/browse/SUP-48894?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/top-bar/top-bar.tsx
+++ b/src/components/top-bar/top-bar.tsx
@@ -1,7 +1,11 @@
 import style from '../../styles/style.scss';
-import {h, Component, VNode} from 'preact';
+import {h, Component, VNode, createRef, RefObject} from 'preact';
 import {connect} from 'react-redux';
 import {PlayerArea} from '../../components/player-area';
+import {bindActions} from '../../utils';
+import {actions} from '../../reducers/shell';
+import {withPlayer} from '../player';
+import {withEventManager} from '../../event';
 
 /**
  * mapping state to props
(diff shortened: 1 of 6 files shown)


src/components/top-bar/top-bar.tsx
```diff
@@ -22,8 +26,19 @@ const COMPONENT_NAME = 'TopBar';
...
+@withPlayer
+@withEventManager
+@connect(mapStateToProps, bindActions(actions))
 class TopBar extends Component<any, any> {
+  topBarRef: RefObject<HTMLDivElement> = createRef<HTMLDivElement>();
+
+  componentDidMount() {
+    const {player, eventManager} = this.props;
+    eventManager.listen(player, player.Event.UI.USER_OPEN_DROPDOWN, () => {
+      this.props.updateTopBarClientRect(this.topBarRef.current?.getBoundingClientRect());
```
SivanA-Kaltura, line 38: now that you've checked that this works - can we listen to an earlier event or to an earlier state change in order to know that we can now compute top bar size ?
is the height on componentDidMount already the full height or not yet ?

SivanA-Kaltura (changes requested):

src/components/top-bar/top-bar.tsx
Tzipi-kaltura: I already checked if we have this in componentDidMount but no we don't have the sizes yet

PR #1096, opened 2025-11-11
feat(FEC-14691):adding an option to display entry title on upper bar

ravitshalem:

Add configuration to display entry name on the upper bar

### DIFF
--- a/src/ui-presets/playback.tsx
+++ b/src/ui-presets/playback.tsx
@@ -31,7 +31,6 @@ import {withKeyboardEvent} from '../components';
 import {VideoArea} from '../components';
 import {GuiArea} from '../components';
 import {ClosedCaptions} from '../components';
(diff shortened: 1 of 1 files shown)


src/components/title/title.tsx
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  const textRef = useRef<HTMLSpanElement | null>(null);
+  const comparisonTextRef = useRef<HTMLSpanElement | null>(null);
+  const [showTooltip, setShowTooltip] = useState(false);
+  const [isFinalized, setIsFinalized] = useState(false);
+  const [width, setWidth] = useState(0);
+
+  let title = '';
+  try {
```
SivanA-Kaltura: you don't need a try catch if you already test that the value is not empty

PR #1163, opened 2026-05-05
fix(SUP-51711): Audio Description Language selection not working as expected in some browsers

SivanA-Kaltura:

Fix audio description menu not being updated in case tracks changed event fires but audio track changed event doesn't fire

Resolves SUP-51711

### DIFF
--- a/src/components/engine-connector/engine-connector.tsx
+++ b/src/components/engine-connector/engine-connector.tsx
@@ -233,7 +233,7 @@ class EngineConnector extends Component<EngineConnectorProps, any> {
       this.props.updateIsCaptionsEnabled(tracks.find(track => track.active)?.language !== 'off');
     });
 
-    eventManager.listen(player, player.Event.Core.AUDIO_TRACK_CHANGED, () => updateAudioTracks);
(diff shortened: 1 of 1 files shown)


src/components/engine-connector/engine-connector.tsx
```diff
@@ -201,13 +219,12 @@ class EngineConnector extends Component<EngineConnectorProps, any> {
...
 
     eventManager.listen(player, player.Event.Core.TRACKS_CHANGED, () => {
-      const audioTracks = player.getTracks(TrackType.AUDIO);
       const videoTracks = player.getTracks(TrackType.VIDEO);
       const textTracks = player.getTracks(TrackType.TEXT);
 
-      this.props.updateAudioTracks(audioTracks.filter(t => !isAudioDescriptionLanguageKey(t.language)));
       this.props.updateVideoTracks(videoTracks);
       this.props.updateTextTracks(textTracks);
+      updateAudioTracks();
```
roeedean, line 227: So the change is to call the function on all `audioTracks` instead of only call on `audioTracks.filter(t => !isAudioDescriptionLanguageKey(t.language))` ?

roeedean (changes requested):

roeedean (approved):

PR #1167, opened 2026-05-13
fix(tooltip): skip showTooltip on programmatic focus (SUP-52316)

MosheMaorKaltura:
## Summary

- **Bug**: Settings gear tooltip becomes permanently visible after pressing ESC to close the settings panel. ESC triggers `button.focus()` on the gear icon (accessibility behaviour), which fires the Tooltip's `onFocus` handler → `showTooltip()`. Since the mouse is not over the button, no `mouseleave` fires, leaving the tooltip stuck until the user moves the mouse.
- **Root cause**: `handleFocusOnChildren()` called `showTooltip()` unconditionally on any focus event, including programmatic `button.focus()` calls introduced as a side-effect of PR #1078 (ADA-1623 ref-wiring fix).
- **Fix**: Guard on `event.relatedTarget`. When it is `null`, focus arrived programmatically (not via keyboard Tab) — skip `showTooltip()`. When it is a real element, the user tabbed to the button — show the tooltip as before. The child's `onFocus` callback is forwarded in both paths so accessibility consumers are not broken.

## Changes

| File | Change |
|---|---|
| `src/components/tooltip/tooltip.tsx` | `handleFocusOnChildren()` now accepts `FocusEvent`; adds `relatedTarget === null` guard |
| `tests/e2e/regression-SUP-52316.spec.js` | New regression test (5 cases: unit guard + integration smoke) |

## Verification

1. ESC closes settings panel → tooltip does **not** appear on gear icon
2. Tab key navigating to gear icon → tooltip **does** appear (relatedTarget is previous element)
3. Mouse hover on gear icon → tooltip **does** appear (unchanged path via `onMouseOver`)

## Test Results

All 16 tests pass (11 pre-existing + 5 new regression tests).

Fixes SUP-52316


MosheMaorKaltura:
## Updated fix: data attribute flag approach (replaces relatedTarget.closest guard)

### Why the previous fixes failed

**Attempt 1 (`relatedTarget.closest('.playkit-control-settings')`):** The settings panel is already unmounted by the time `focusElement()`'s `setInterval` fires `.focus()` ~100ms after `componentWillUnmount`. At that point `relatedTarget` is `document.body` — the panel element is gone from the DOM so `closest()` finds nothing.

**Attempt 2 (`skipFocusRestore` prop):** The prop is set to `true` in the same render cycle that unmounts the panel. `componentWillUnmount` runs with the **old** props (before the update), so `skipFocusRestore` is still `false` when the guard is checked.

### The correct fix

**`src/utils/popup-keyboard-accessibility.tsx` — `componentWillUnmount`:**
Set `data-kaltura-focus-restore` on the trigger element **synchronously**, before `focusElement()` starts its interval. This is guaranteed to be on the element when the interval eventually fires `.focus()`:

```diff
 componentWillUnmount(): void {
-  if (this.props.skipFocusRestore) return;
   if (this._previouslyActiveElement && document.contains(this._previouslyActiveElement) && !this.isOverlayOpen) {
+    (this._previouslyActiveElement as HTMLElement).dataset.kalturaFocusRestore = 'true';
     focusElement(this._previouslyActiveElement);
```

**`src/components/tooltip/tooltip.tsx` — `handleFocusOnChildren`:**
Check for the flag first. If present, consume it and return without calling `showTooltip()`:

```diff
 handleFocusOnChildren = (event: FocusEvent): void => {
   const {onFocus} = (this.props.children as VNode<any>).props;
+  const target = event.target as HTMLElement;
+  if (target?.dataset?.kalturaFocusRestore) {
+    delete target.dataset.kalturaFocusRestore;
+    if (onFocus) onFocus(event);
+    return;
+  }
   this.showTooltip();
```

The previous `relatedTarget.closest()` guard is removed entirely and replaced with this approach.

### Why this works

- The flag is set **synchronously** in `componentWillUnmount` — before `setInterval` fires
- When `focus()` is called ~100ms later, `handleFocusOnChildren` sees the flag and skips `showTooltip()`
- The flag is deleted after being consumed so it doesn't persist
- Normal user focus (Tab key, mouse click) does not set the flag so tooltip behaves normally

src/utils/popup-keyboard-accessibility.tsx
```diff
@@ -113,6 +113,7 @@ export const withKeyboardA11y = (WrappedComponent): any => {
      */
     componentWillUnmount(): void {
       if (this._previouslyActiveElement && document.contains(this._previouslyActiveElement) && !this.isOverlayOpen) {
+        (this._previouslyActiveElement as HTMLElement).dataset.kalturaFocusRestore = 'true';
```
MosheMaorKaltura: use true as boolean and not as string

lianbenjamin (approved):

PR #1172, opened 2026-05-18
fix(ADA-3347): Add aria-label to landmark region

valentinStoicaK:
This is fixing https://kaltura.atlassian.net/browse/ADA-3347

**Issue:**

**Fix:**

### DIFF
--- a/src/components/video-player/video-player.tsx
+++ b/src/components/video-player/video-player.tsx
@@ -17,7 +17,7 @@ import {PlayerArea} from '../player-area';
 const mapStateToProps = state => ({
   videoStyles: state.shell.layoutStyles.video,
   targetId: state.config.targetId,
-  sources: state.engine.sources
+  entryName: state.engine.sources?.metadata?.name
 });
 
 /**
@@ -45,7 +45,7 @@ class VideoPlayer extends Component<any, any> {
    * @memberof VideoPlayer
    */
   shouldComponentUpdate(nextProps: any): boolean {
-    return nextProps.videoStyles !== this.props.videoStyles || nextProps.sources !== this.props.sources;
+    return nextProps.videoStyles !== this.props.videoStyles || nextProps.entryName !== this.props.entryName;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/components/video-player/video-player.tsx
```diff
@@ -40,7 +45,7 @@ class VideoPlayer extends Component<any, any> {
    * @memberof VideoPlayer
    */
   shouldComponentUpdate(nextProps: any): boolean {
-    return nextProps.videoStyles !== this.props.videoStyles;
+    return nextProps.videoStyles !== this.props.videoStyles || nextProps.sources !== this.props.sources;
```
ravitshalem: Why checking the entire sources object?
it has lots of fields inside. Even if something unrelated inside sources changes, the component will Unnecessarily re-render. 
 instead why not check the specific field it actually uses? for exmaple sources.metadata.name.

src/components/video-player/video-player.tsx
```diff
@@ -15,7 +16,8 @@ import {PlayerArea} from '../player-area';
  */
 const mapStateToProps = state => ({
   videoStyles: state.shell.layoutStyles.video,
-  targetId: state.config.targetId
+  targetId: state.config.targetId,
+  sources: state.engine.sources
```
ravitshalem: you pulls the entire sources object from the Redux store, but only ever uses sources.metadata.name, is it really needed?

valentinStoicaK: You are right, I should use only what is needed. Fixed!

roeedean (dismissed):

src/components/video-player/video-player.tsx
```diff
@@ -25,9 +27,12 @@ const mapStateToProps = state => ({
  * @example <VideoPlayer>...</VideoPlayer>
  * @extends {Component}
  */
+@connect(mapStateToProps, bindActions(actions))
```
ravitshalem: is there any reason the connect decorator moved to the top?
originally is was under withEventManager decorator,
since the decorators are applied bottom-top, this change can cause changes

valentinStoicaK: When adding the @withText decorator I noticed this is the usual order in this whole repo, that's why I changed it. Do you think I should revert this change?

ravitshalem: from Claude's review:
"The @connect decorator was moved from innermost to outermost, which is a non-trivial behavioral change unrelated to the accessibility fix.
@withText doesn't need to be in any specific position relative to @connect — it uses IntlProvider context, not Redux. The minimal, safe change would preserve the original decorator order and just insert @withText at the bottom."

can you check it?

ravitshalem (approved):

roeedean (approved):

PR #1177, opened 2026-06-02
fix(FEC-14971): hide audio menu when only one meaningful audio track …

MosheMaorKaltura:
…exists

DASH streams with HD flavors produce extra audio tracks with label:null, resolved to language:'und'. Filter these out before updating the store so the audio button and settings menu stay hidden for single-language content.

**Issue:**

**Fix:**


src/components/engine-connector/engine-connector.tsx
ravitshalem, line 77: audioTracks is also being Passed on updateDefaultAdvancedAudioDescription
 do we need to pass here namedTracks as well?

ravitshalem (approved):

PR #1188, opened 2026-07-29
fix(FEC-15048): Add play pause keyboard shortcut

valentinStoicaK:
https://kaltura.atlassian.net/browse/FEC-15048
Adds K key as a keyboard shortcut for Play/Pause button. The shortcut is bypassed when focus is on the pre-playback play button, so that he behavior is not changed.


src/components/shell/shell.tsx
```diff
@@ -84,6 +86,55 @@ class Shell extends Component<any, any> {
...
+    }
+
+    //Prevents keyboard shortcuts from firing when the user is interacting with the audio description
+    const isInsideAudioDescriptionRoot = (!!targetElement && !!targetElement.closest('.aadRoot')) || (!!activeHTMLElement && !!activeHTMLElement.closest('.aadRoot'));
+    if (isInsideAudioDescriptionRoot) {
+      return;
+    }
+
+    // Preserve pre-playback overlay keyboard UX by letting its focused button handler own the interaction.
+    if (this.props.prePlayback) {
```
roeedean, line 111: the description says the shortcut is "bypassed when focus is on the pre-playback play button".
The code bypasses when `this.props.prePlayback` (Redux) is truthy - that's the pre-playback state, not the pre-playback button's focus. If a user tabs elsewhere while pre-playback is showing, `K` does nothing. Confirm the intent.

It is probably fine, but the description is misleading.

src/components/shell/shell.tsx
```diff
@@ -84,6 +86,55 @@ class Shell extends Component<any, any> {
...
+
+    if ((!this.props.player.isLive() && !this.props.allowPlayPause) || (this.props.player.isLive() && !this.props.allowLivePlayPause)) {
+      return;
+    }
+
+    const engineState = this.props.player?.ui?.store?.getState()?.engine;
+    const isPlayingFromStore = engineState ? (engineState.adBreak ? engineState.adIsPlaying : engineState.isPlaying) : this.props.isPlayingAdOrPlayback;
+    e.preventDefault();
+    e.stopPropagation();
+    isPlayingFromStore ? this.props.player.pause() : this.props.player.play();
```
roeedean: `K` bypasses the `notifyClick` / UI-event pipeline - regression vs. `SPACE`.

The play/pause button and the `SPACE` keyboard handler both go through `notifyClick()` ->> event-dispatcher middleware ->> `onPlayPauseClicked`, which dispatches the public `USER_CLICKED_PLAY` / `USER_CLICKED_PAUSE` events.

The new `Shell.onDocumentKeyDownCapture` calls `player.play()` / `player.pause()` directly and skips that entire pipeline, so those events are never dispatched on `K`. Integrations that count user-initiated play/pause will hear `SPACE` and button clicks but stay silent on `K` - a behavior split within the same semantic action.

Side effect: the `[KeyMap.K]` entries added to `keyboardHandlers` in `event-dispatcher.ts` and `event-dispatcher-provider.tsx` are currently dead code, since nothing dispatches a `notifyClick({key: KeyMap.K, ...}`).

Suggested preferred fix (by Claude): wire `K` into the Keyboard component next to `SPACE` (`keyboard.tsx:112`) so it flows through `notifyClick` ->> middleware.
The `keyboardHandlers[KeyMap.K]` entries you already added then start doing real work, and the extra document-level listener in Shell can be removed. `K` stays on the exact same code path as `SPACE` - one place to change, one place to test.

Alternative (by Claude): keep the Shell listener but replace the direct `player.play()` / `player.pause()` calls with `player.dispatchEvent(new FakeEvent(EventType.USER_CLICKED_PLAY | USER_CLICKED_PAUSE))` (same as `onPlayPauseClicked`), and drop the unused `KeyMap.K` middleware entries.

roeedean (changes requested):

src/components/shell/shell.tsx
valentinStoicaK: I will go with the alternative fix instead. The SPACE key shortcut is not working in the first place. I can see that the Keyboard component is not even used. I am keeping the Shell keydown listener and adding the USER_CLICKED_PAUSE and USER_CLICKED_PLAY before toggling playback

valentinStoicaK:
@ravitshalem the comments based on Claude's review:
1. event-dispatcher.ts change is dead code - Removed from here;
2. Bypasses the established keyboard infrastructure - SPACE is not using KeyboardEventProvider → withKeyboardEvent HOC as Claude is suggesting, but the Keyboard component that is dead code. Routing K throught it would wire it into an infrastructure that is not functional today;
3. Missing overlay icon update - Done;
4. Triple key detection is redundant - Done;
5. Fragile store access when this.props already has the value - Done;
6. prePlayback guard makes K inert on the pre-playback screen unless button is focused - This is intended behavior;
7. PR description still contains template placeholders - Done;
8. pre-playback-play-overlay.tsx missing e.repeat guard - I don't think this is needed. Pre-playback play button can be actioned only once;

roeedean (approved):

PR #1191, opened 2026-08-07
fix(ADA-3372): Fix invalid ARIA structure in error overlay

valentinStoicaK:
This fixes https://kaltura.atlassian.net/browse/ADA-3372 by changing role alert to role alertdialog to be valid in combination with aria-modal.
Also this handles the focus when error overlay appears.

### DIFF
--- a/src/components/error-overlay/error-overlay.tsx
+++ b/src/components/error-overlay/error-overlay.tsx
@@ -11,6 +11,7 @@ import {withPlayer} from '../../components/player';
 import {Button} from '../../components/button';
 import {getErrorDetailsByCategory} from './error-message-provider';
 import {actions as overlayActions} from '../../reducers/overlay';
+import {getPlayerHadFocusOnError} from '../engine-connector/engine-connector';
 
 /**
  * mapping state to props
@@ -37,12 +38,30 @@ const COMPONENT_NAME = 'ErrorOverlay';
 @withLogger(COMPONENT_NAME)
 class ErrorOverlay extends Component<any, any> {
   private sessionEl!: HTMLDivElement;
+  private overlayContentRef: HTMLDivElement | null = null;
 
   constructor(props: any) {
     super(props);
     this.state = {
       entryUrl: props.componentData?.errorOverlay || undefined
     };
+  }
+
+  public componentDidMount(): void {
+    // Check if THIS player had focus when error occurred
+    const targetId = this.props.player?.config?.targetId;
+    if (targetId) {
+      const hadFocus = getPlayerHadFocusOnError(targetId);
+      if (hadFocus && this.overlayContentRef) {
+        // Try to focus first interactive element, fallback to container
+        const focusable = this.overlayContentRef.querySelector<HTMLElement>('button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])');
+        if (focusable) {
+          focusable.focus();
+        } else {
+          this.overlayContentRef.focus();
+        }
+      }
+    }
   }
 
   public componentDidUpdate(prevProps: any): void {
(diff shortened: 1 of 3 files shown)


src/components/error-overlay/error-overlay.tsx
```diff
@@ -149,8 +168,8 @@ class ErrorOverlay extends Component<any, any> {
 
     return (
       <div className={style.headline}>
-        <div className={style.errorTitle}>{this.props.errorHead || errorTitleRes}</div>
-        {errorMessageRes ? <div className={style.errorMessage}>{errorMessageRes}</div> : undefined}
+        <div id="error-overlay-title" className={style.errorTitle}>{this.props.errorHead || errorTitleRes}</div>
```
roeedean: `id="error-overlay-title"` static string, we already supports multiple player instances on a single page and have a standard way of scoping IDs by `targetId` (e.g., video-player.tsx:121: `id={`${targetId}-video`}`).

With two players, you get duplicate IDs which is invalid, and `aria-labelledby/aria-describedby` resolve to the first match in document order - the second player's overlay announces the first player's text.

Fix: scope the ID (`id={${targetId}-error-overlay-title}`), and pass the composed values to `ariaLabelledBy`/`ariaDescribedBy`.

src/components/error-overlay/error-overlay.tsx
```diff
@@ -149,8 +168,8 @@ class ErrorOverlay extends Component<any, any> {
...
+        {errorMessageRes ? <div id="error-overlay-message" className={style.errorMessage}>{errorMessageRes}</div> : undefined}
```
roeedean: `id="error-overlay-message"` static string, we already supports multiple player instances on a single page and have a standard way of scoping IDs by `targetId` (e.g., video-player.tsx:121: `id={`${targetId}-video`}`).

With two players, you get duplicate IDs which is invalid, and `aria-labelledby/aria-describedby` resolve to the first match in document order - the second player's overlay announces the first player's text.

Fix: scope the ID (`id={${targetId}-error-overlay-message}`), and pass the composed values to `ariaLabelledBy`/`ariaDescribedBy`.

src/components/error-overlay/error-overlay.tsx
```diff
@@ -167,8 +186,8 @@ class ErrorOverlay extends Component<any, any> {
       const errorOverlayStyles = backgroundUrl ? {backgroundImage: `url(${backgroundUrl})`} : undefined;
       return (
         <div className={['overlay-portal', backgroundUrl ? style.customErrorSlate : ''].join(' ')}>
-          <Overlay open permanent={true} type="error">
-            <div className={style.errorOverlay} style={errorOverlayStyles}>
+          <Overlay open permanent={true} type="error" ariaLabelledBy="error-overlay-title" ariaDescribedBy="error-overlay-message">
```
roeedean: Fix: pass the composed values to `ariaLabelledBy`/`ariaDescribedBy`.

`{${targetId}-error-overlay-title}`
`{${targetId}-error-overlay-message}`

src/components/overlay/overlay.tsx
roeedean: `aria-modal="true"` is hardcoded on the overlay div, present on all overlay types including error.
This PR changes the role on that same element from `alert` to `alertdialog` but doesn't touch `aria-modal`.

Conditional focus conflicts with `aria-modal`. `aria-modal="true"` signals that the rest of the page is inert.
If focus isn't moved inside (when `hadFocus === false`), you have a modal claiming to trap interaction while focus remains outside.

roeedean (changes requested):

src/components/overlay/overlay.tsx
valentinStoicaK: Having in mind that all overlays are player-scoped (not page level), none of the overlays should have aria-modal="true". Removing it entirely

roeedean (approved):

PR #1194, opened 2026-08-13
fix(ADA-3480): Add title to Settings close button

valentinStoicaK:
This fixes https://kaltura.atlassian.net/browse/ADA-3480 by adding title to close button aria-label.

### DIFF
--- a/src/components/settings/settings.tsx
+++ b/src/components/settings/settings.tsx
@@ -47,6 +47,7 @@ const COMPONENT_NAME = 'Settings';
 
 const translates = () => ({
   buttonLabel: <Text id="controls.settings">Settings</Text>,
+  settingsText: 'controls.settings',
   qualityHdLabel: <Text id="settings.qualityHdLabel">Quality is HD</Text>,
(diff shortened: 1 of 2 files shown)


src/components/settings/settings.tsx
```diff
@@ -47,6 +47,7 @@ const COMPONENT_NAME = 'Settings';
 
 const translates = () => ({
   buttonLabel: <Text id="controls.settings">Settings</Text>,
+  settingsText: 'controls.settings',
```
roeedean, line 50: Why this line added differently from the others?

What you added:
- `settingsText: 'controls.settings',`

What is already there:
- `buttonLabel: <Text id="controls.settings">Settings</Text>,`
- `qualityHdLabel: <Text id="settings.qualityHdLabel">Quality is HD</Text>,`

roeedean (changes requested):

src/components/settings/settings.tsx
valentinStoicaK: The buttonLabel is JSX. I needed a string for closeAriaLabel so I can concatenate it with close text.

roeedean (approved):

PR #1198, opened 2026-08-25
fix(FEC-15083): Add right controls to xs player

valentinStoicaK:
This fixes https://kaltura.atlassian.net/browse/FEC-15083

### DIFF
--- a/src/components/bottom-bar/bottom-bar-utils.ts
+++ b/src/components/bottom-bar/bottom-bar-utils.ts
@@ -5,7 +5,11 @@ export function filterControlsByPriority(
   currentControlWidth: number,
   lowerPriorityControls: string[][]
 ): string[] {
-  const numOfOverflowControls = Math.ceil((currentMinBreakPointWidth - currentBarWidth) / currentControlWidth) || 1;
+  // Add 20% buffer to account for spacing/margins between controls
+  const overflow = currentMinBreakPointWidth - currentBarWidth;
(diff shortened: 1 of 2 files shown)


src/components/bottom-bar/bottom-bar-utils.ts
```diff
@@ -5,7 +5,11 @@ export function filterControlsByPriority(
   currentControlWidth: number,
   lowerPriorityControls: string[][]
 ): string[] {
-  const numOfOverflowControls = Math.ceil((currentMinBreakPointWidth - currentBarWidth) / currentControlWidth) || 1;
+  // Add 20% buffer to account for spacing/margins between controls
```
ravitshalem, line 8: Why specifically 20% was chosen?
Is it possible that for player with many controls it will cause one of them to disappear?

roeedean (approved):

src/components/bottom-bar/bottom-bar-utils.ts
valentinStoicaK: This is what I'm trying to achieve. After changing the display to flex for this version, the right controls wrap on a new row when there is not enough space on xs players. Now, if there is not enough space, one more control is removed. This is based on the LOWER_PRIORITY_CONTROLS array. 
The 20% is trying to add to the numOfOverflowControls calculation the width of the gap here 
image

ravitshalem (approved):

